PR #12, opened 2021-01-06
Beef 2.0 I guess?

GuillaumePrata:
## About The Pull Request
Adds Beefman and it kinda works...
There are still 2 issues to iron out that I will try tomorrow if real life let me. Else they still work and someone can fix it in the future.
1- They are faceless, the eyes and mouth sprites aren't working, they do look extra spooky tho.
2- Sometimes your dismembered limb don't turn into meat and instead drop a limb itself which you can't easily mash back into you.

## Why It's Good For The Game
It isn't, beefman bad and spooky!

## Changelog
:cl:
add: Added Beefman as a species to play again!
tweak: You now need harm intent to re-arm yourself
balance: They don't bleedout at room temperature anymore but they still bleed out and die in a minute if something like a fever happens, likely will need harsher drawbacks to balance it out in the future.
/:cl:


JohnFulpWillard:
Why is this also porting over Bloodsucker stuff? Would that cause merge conflicts for the bloodsucker branch?

GuillaumePrata:
It was only porting icons and a define that I forgot because I didn't expect the linter error and me not having time to figure out how to fix it, still I can purge everything properly when I finish this... today maybe...

GuillaumePrata:
Actually I will restart this PR to because it is easier to just add things instead of rewriting everything again, be right back!

PR #37, opened 2021-01-29
Fix compile errors

remove32:
## About The Pull Request
fixes some busted variables

## Why It's Good For The Game
not broken = good


remove32:
idk why it looks like that when all i did was the most recent commit what whatever, christ, git is finicky sometimes

PR #41, opened 2021-02-10
Adds random job option

GoodDays13:
## About The Pull Request

Adds a random job option in the late join menu that chooses a random job that is available and lets you either choose it, reroll, or just close it and choose one yourself instead. (This is my first thing made in DM and it was bug tested by me, so I may have missed something.)

## Why It's Good For The Game

When joining, I always have trouble choosing what job to be, so I always choose the same job. If this option was here, I would get more experience with other jobs. It chooses the job for me.

## Changelog
:cl:
add: Added a "Random" option in the job choice menu when joining an in-progress round
/:cl:

### DIFF
--- a/code/modules/mob/dead/new_player/new_player.dm
+++ b/code/modules/mob/dead/new_player/new_player.dm
@@ -175,7 +175,28 @@
 	if(href_list["manifest"])
 		ViewManifest()
 
+	if(href_list["cancrand"])
+		src << browse(null, "window=randjob") //closes the random job window
+		LateChoices()
+		return
+
 	if(href_list["SelectedJob"])
+		if(href_list["SelectedJob"] == "Random")
+			var/list/dept_dat = list()
+			for(var/category in GLOB.position_categories)
+				for(var/job in GLOB.position_categories[category]["jobs"])
+					var/datum/job/jobs = SSjob.name_occupations[job]
+					if(jobs && IsJobUnavailable(jobs.title, TRUE) == JOB_AVAILABLE)
+						dept_dat += jobs.title
+			var/rando = dept_dat[rand(1, dept_dat.len)]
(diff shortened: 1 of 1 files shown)


code/modules/mob/dead/new_player/new_player.dm
```diff
@@ -175,7 +175,28 @@
...
+			var/randomjob = "<p><center><a href='byond://?src=[REF(src)];SelectedJob=[rando]'>[rando]</a></center><center><a href='byond://?src=[REF(src)];SelectedJob=Random'>Reroll</a></center><center><a href='byond://?src=[REF(src)];cancrand=[1]'>Cancel</a></center></p>"
+
+			var/datum/browser/popup = new(src, "randjob", "<div align='center'>Random Job</div>", 200)
+			popup.set_window_options("can_close=0")
+			popup.set_content(randomjob)
+			popup.open(FALSE)
+			return
+
 		if(!SSticker?.IsRoundInProgress())
 			to_chat(usr, "<span class='danger'>The round is either not ready, or has already finished...</span>")
```
Fikou, line 201: cant you abuse the random job button if it runs before the round in progress thing

GoodDays13: I mean, the code that happens before the round in progress thing just opens a window that lets you choose a job, but choosing the job it shows should do the same thing as picking it from the normal late choices menu.

Fikou: ah ok

remove32:
how is this any different from just having everything at the same level?

GoodDays13:
> would this somehow bypass role restrictions and role bans?

I don't think so, and I just tested it and it seemed to work fine for me, banned jobs never came up as an option. I'm not sure what you're saying in the first sentence.

BootlegBow:
> how is this any different from just having everything at the same level? 

as far as i can tell, this is a button you press when latejoining which gives you a random job, it's not a pre-round preference set in character creation

Helianthus00:
This touches /tg/ files and isn't a significant feature. I suggest you try PRing it there.

GoodDays13:
eh, isn't worth it. i don't care enough

PR #45, opened 2021-02-19
build.bat will no longer skip fulp_modules

SgtHunk:
You know what? fuck you. *skips your fulp_modules folder*

## About The Pull Request

Makes fulp_modules part of dependencies so that it doesn't get skipped in case a change is made in ONLY that folder.

## Why It's Good For The Game

I've had multiple breakdowns but it doesn't compare to the ones i've had when I see 'skipping DM' only because I edited a file in there without editing any other file.

### DIFF
--- a/tools/build/build.js
+++ b/tools/build/build.js
@@ -32,6 +32,9 @@ const taskTgui = new Task('tgui')
 
 const taskDm = new Task('dm')
   .depends('code/**')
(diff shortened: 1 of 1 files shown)


JohnFulpWillard:
fuck build.bat, eat lead

remove32:
good change. good hunk

PR #50, opened 2021-02-23
Engineering Department Tablets

urococt:
## About The Pull Request

Replaces the tablets given to station engineers, atmospheric technicians, and chief engineers with reskinned versions that have appropriate software installed by default.

## Why It's Good For The Game

Waiting for pretend software to download is not fun. The new sprites add visual flavor.

## Changelog
:cl:
add: Tablets issued to the engineering department now come with useful software pre-installed and have fancy new paint jobs
/:cl:


GuillaumePrata:
While the idea is interesting, this overwrites some outfits we already have in game and also change the icon file.
What would happen is that every time a new PR change anything on the CE/SE/AT outfits we would need to rewrite your overwrite and every time a new sprite is added to the modular tablets we would need to update the icon file or we would have broken icons.

All this extra work just so people don't need to spend a few seconds at the start of the round downloading files simply isn't worth.

urococt:
> What would happen is that every time a new PR change anything on the CE/SE/AT outfits we would need to rewrite your overwrite

The way the outfit changes are written (although I think it breaks style guidelines), they would only have to be updated if backpack_contents is changed.

> every time a new sprite is added to the modular tablets we would need to update the icon file or we would have broken icons.

You make a good point. If this is a dealbreaker, I'm not opposed to removing the sprite changes. I would even be willing to only add software to the atmos tablet, since that's already its own class.

Helianthus00:
This seems to be an unnecessary change just to save a small bit of time.

PR #51, opened 2021-02-23
I dont like Mentors

JohnFulpWillard:
## About The Pull Request

The PR should (I hope) all good and done, ready for merging.

What does this PR change, exactly?
- The color for fellow Mentors seeing what a Mentor has responded to someone, has been changed from Blue to Purple, because it was an eyesore to look at
- Mentorhelp/MentorSay are under the Mod chat in the chat filter, rather than being left unfiltered, the exception to this is responses non-mentors get (you signed up to get a mentor response, you will get one whether you like it or not. Also we had a few problems of people turning mod chat off for some reason, then said they couldn't see mentorhelps)
- You can now cancel out of Mentorsays
- The Mentorsay verb is back in the Mentor tab (as long as you're a Mentor+)
- The never-to-be-used configs for Mentors were removed
- The timer for Mhelping things has been fixed
- The in-code comments were made cleaner (to me, anyway) and was made less bad


Helianthus00:
Seems to work as intended. Also, mod chat does exist in the chat tabs option. This will be very helpful to group in with asay to see them all at once.

PR #71, opened 2021-03-15
Test time

GuillaumePrata:
Just death


EgorDinamit:
Why. You can just disable Families in config for dynamic, can't you?

GuillaumePrata:
> Why. You can just disable Families in config for dynamic, can't you?

Dynamic 2021 wasn't compatible with the json
To quote the PR itself

> To server owners: this is incompatible with current dynamic.json configurations. This is not just because some fields have different values now, but because the older values are not balanced towards this system.

PR #72, opened 2021-03-15
Cult stun nerf & Cult blind buff

JohnFulpWillard:
## About The Pull Request

Nerfs the Cult stun to be less effective on Mindshielded folk, and gives SecHUDs to the Zealot's blindfold.

## Why It's Good For The Game

Mindshields are meant to be an important counter to conversion antagonists (Revs, Cult, Bloodsuckers), but this is ignored on Cultists. This is because Mindshields give no benefit, and the only difference is that it permanently removes you from the round instead of getting converted and at least having a chance at getting deconverted later. This causes the crew to actively avoid getting mindshielded, and things like these to happen:
image

## Changelog
:cl: John Willard
balance: The Cult's stun arm is less effective on Mindshielded folk
balance: The Cult's blindfold now shows who is Mindshielded
/:cl:


SgtHunk:
You currently have a negative Fix/Feature pull request delta of -6. Maintainers may close this PR at will. Fixing issues or improving the codebase will improve this score.

JohnFulpWillard:
> 
> 
> You currently have a negative Fix/Feature pull request delta of -6. Maintainers may close this PR at will. Fixing issues or improving the codebase will improve this score.

You have a negative Fuck/You your mom delta of -6. Maintainers may close this YM at will. Fixing issues or improving the codebase will improve this score.

EgorDinamit:
> You currently have a negative Fix/Feature pull request delta of -6. Maintainers may close this PR at will. Fixing issues or improving the codebase will improve this score.

fix: Fixes balance around the cult stun and cult blindfold.

fulp_modules/cult_changes/cult_changes.dm
```diff
@@ -0,0 +1,74 @@
...
+
+		else
+			if(HAS_TRAIT(target, TRAIT_MINDSHIELD))
+				var/mob/living/carbon/C = L
+				to_chat(user, "<span class='cultitalic'>Their mind is too strong, resisting the spell, but it damaged them nonetheless!</span>")
+				C.stuttering += 8
+				C.dizziness += 20
+				C.Jitter(6)
+				C.bleed(30)
+				C.apply_damage(60, STAMINA, BODY_ZONE_CHEST)
```
necromanceranne, line 38: One of the major issues with stamina damage for cultists is that they generally don't have much of it at all. Typically, when they run into mindshielded people, their last resort is lethal attacks which is all brute and the like. This was the major problem with this same change on tg, where baton jousts ended up favouring anyone with a baton due to their followup on the first attack, which stun doesn't have given it expends the attack immediately after use.

Since this is causing a bleed effect, I might recommend instead having this deal a (reasonable) amount of brute and have wounding power instead, and maybe a knockdown to top it off. Once everyone is mindshield, it becomes incredibly easy to streamroll into a cult base with your effect new brain armor, and cultists can't differentiate a mindshielded target from a nonmindshelded target.

JohnFulpWillard: > One of the major issues with stamina damage for cultists is that they generally don't have much of it at all. 

We dont want them using Stamina damage against Mindshielded people, afterall mindshielded folk gets their body permanently removed from the round if they lose.

> their last resort is lethal attacks which is all brute and the like.

Lethals on Mindshielded should be their first resort, the stamina damage is just to slow the user down so they can't escape at least, but maybe I can lower it.

> This was the major problem with this same change on tg, where baton jousts ended up favouring anyone with a baton due to their followup on the first attack, which stun doesn't have given it expends the attack immediately after use.

This is a fair point, I might remove the stamina damage entirely afterall.

> and cultists can't differentiate a mindshielded target from a nonmindshelded target.

Thats why I added sechuds into the zealot's

Helianthus00:
It's unfortunate that mindshielded people aren't easy to convert but the metagaming of not using mindshields is more of a policy issue than a coding one. This isn't really worth editing tg files as far as I can see.

PR #75, opened 2021-03-20
Makes the Radioactive mutation ethereal only

FernandoJ8:
## About The Pull Request

This PR makes it so that the Radioactive mutation can only be obtained by Ethereals in the same fashion as Lizardpeople's fire breath.

## Why It's Good For The Game

Hulk spam and hulk-tiding is a common enough problem that something should be done about it. This PR makes Hulk more difficult to obtain and encourages geneticists to actively collaborate with other crew members.

## Changelog
:cl:
balance: Made the Radioactive mutation only obtainable from Ethereals
/:cl:


ATH1909:
> This PR makes it so that the Radioactive mutation can only be obtained by Ethereals in the same fashion as Lizardpeople's fire breath.

okay, how are you going to tell people about this? this is pretty important for a geneticist who's using the /tg/ wiki to know.

remember that shock touch is also locked behind radioactive.

FernandoJ8:
> okay, how are you going to tell people about this? this is pretty important for a geneticist who's using the /tg/ wiki to know.
> 
> remember that shock touch is also locked behind radioactive.

As for shock touch, I think it's appropriate that it's also locked behind the race that's all about electricity.

Regarding the way of getting this info around, I'd say adding a page to the fulp wiki with miscellaneous would probably be a good thing in the long run. Activity here is rising and there will likely be an increase in fulp only changes that are too small or quick to explain to merit a page on their own but are still too significant to not mention at all. Willard's Cult Stun rework is a recent example of that as well.

Horatio22:
I'm really not on-board with this one. There are plenty of legitimate reasons why someone would take hulk, and taking that away because people break sever rules with hulk isn't the right way to go, imo. Besides, ethereal hulks make [no sense lore-wise,](https://tgstation13.org/wiki/Ethereal ) and to be frank, a glowing hulk would look kinda strange. Personally, I'd much rather see something in the research tree for sec upgrades that would allow them to better handle hulks than nerf the mutation

FernandoJ8:
> 
> 
> I'm really not on-board with this one. There are plenty of legitimate reasons why someone would take hulk, and taking that away because people break sever rules with hulk isn't the right way to go, imo. Besides, ethereal hulks make [no sense lore-wise,](https://tgstation13.org/wiki/Ethereal) and to be frank, a glowing hulk would look kinda strange. Personally, I'd much rather see something in the research tree for sec upgrades that would allow them to better handle hulks than nerf the mutation
I don't get how I'm taking anything away, all I'm doing is making so that genetics has to find a single ethereal volunteer to get the radiation mutation from.
I don't get what you mean by making no sense lore wise, ethereals may be more pacifistic but that doesn't have anything to do with your genes, and besides ethereals only have the radiation mutation which does make sense for them to have. Mixing it with strenght is what, in-lore, creates the artificial mutation that is Hulk

PR #76, opened 2021-03-22
Nightmares arent smart

JohnFulpWillard:
## About The Pull Request

The server rules prevent the use of nightmares from using advanced machinery. Instead of having this as a rule hidden in the wiki page where people that failed to read it breaks it, its instead in the code, to prevent these problems in the first place.

## Why It's Good For The Game

Less work on the admins

## Changelog
:cl:
balance: Nightmares are no longer intelligent
/:cl:


ATH1909:
> The server rules prevent the use of nightmares from using advanced machinery.

bruh why is that even a rule

EgorDinamit:
> bruh why is that even a rule

Because I once released a singularity as a nightmare.

Horatio22:
> 
> 
> > The server rules prevent the use of nightmares from using advanced machinery.
> 
> bruh why is that even a rule

because they, for some reason or another, would sit in viro for 10 minutes and make super aids, which sucks as a midround antag that has a literal saw for an arm

Likteer:
> sit in viro for 10 minutes and make super aids, which sucks as a midround antag that has a literal saw for an arm

As hilarious as this example sounds to me, I agree.

JohnFulpWillard:
what the absolute shit happened when I updated with TG

PR #80, opened 2021-03-28
[READY] [PORT] Adds Combat Mode

SgtHunk:
Title
actually fixes IAA i think
only for TM purposes


Helianthus00:
speedmerging

SgtHunk:
Worked as intended

PR #104, opened 2021-04-08
TG Update

### LINKED ISSUE #102: Supply shuttle launches with people in it
## Round ID: 
[13774](https://scrubby.melonmesa.com/round/13774)

## Testmerges:
The following pull requests are currently test merged:
#97: 'Bloodsuckers' by JohnFulpWillard at commit abf2cae30b

## Reproduction:
1. Play as deputy, which is what I was at the time. (test other roles if you wish)
2.  Hit the launch button. (Do this after step 3 if there isn't a delay before launching, can't remember if there is)
3. Board the cargo shuttle.
4. Reach centcom and you will be ghosted by the game.

JohnFulpWillard:
## About The Pull Request

TG has some critical bug fixes that we could use ASAP.

Will fix #102 

## Changelog
Not needed, the TGU handles this for us.


JohnFulpWillard:
another epic fail, ill be back maybe

Helianthus00:
I'm going to disable mapdiffbot just in case...

Helianthus00:
Still need to do some other map stuff but there is the bulk of it.

PR #122, opened 2021-04-13
science deputy medipens now contain a sane amount of mutadone

ATH1909:
## About The Pull Request

The mutadone medipen that each Science deputy receives as a part of their roundstart kit now contains 10u of mutadone instead of 60u of mutadone.

Science deputies now start each shift with two mutadone medipens instead of one.

Mutadone medipens now use the sprites for atropine medipens instead of the sprites for epinephrine medipens.

## Why It's Good For The Game

John Willard gave each Science deputy a mutadone medipen as a part of their starting kit. For some ungodly reason, this mutadone medipen contains **60u of mutadone** and injects **30u of mutadone** per jab (so, unlike the epineprhine medipen it shares a sprite with, it actually contains two doses, not one. that definitely won't confuse new players!). For reference, mutadone only needs one tick to purge your mutations, and it'll keep purging your mutations every tick until it leaves your system. In the case of a 30u mutadone dose, the time needed for mutadone to leave your system is TWO AND A HALF MINUTES, which is _gross_ overkill.

The new 10u dose reduces that to a much saner 50 seconds, which should be more than enough time to finish an arrest in. It's also consistent with the dosage volumes of most of the other medipens in the game, including the atropine, oxandrolone, salicylic acid, pentetic acid, and salbutamol medipens. The new dosage also opens up the possibility of mutadone medipens being added to the list of medipens that can be refilled by medipen refillers.

Mutadone medipens should not share sprites with epinephrine medipens, as players can very easily confuse one for the other, especially since the latter is in every job's roundstart kit. Since I'm not a spriter, I decided to switch them over to the atropine medipen sprites, since atropine medipens are also pretty niche and uncommon.

## Changelog
:cl:
balance: The mutadone medipen that each Science deputy receives as a part of their roundstart kit now contains 10u of mutadone instead of 60u of mutadone.
balance: Science deputies now start each shift with two mutadone medipens instead of one.
imageadd: Mutadone medipens now use the sprites for atropine medipens instead of the sprites for epinephrine medipens.
/:cl:

### DIFF
--- a/fulp_modules/jobs/deputy/code/deputy.dm
+++ b/fulp_modules/jobs/deputy/code/deputy.dm
@@ -179,6 +179,9 @@ GLOBAL_LIST_INIT(available_deputy_depts, sortList(list(SEC_DEPT_ENGINEERING, SEC
 /obj/item/reagent_containers/hypospray/medipen/mutadone
 	name = "mutadone medipen"
 	desc = "Contains a chemical that will remove all of an injected target's mutations, including positive ones."
+	icon_state = "atropen"
+	inhand_icon_state = "atropen"
+	base_icon_state = "atropen"
 	volume = 10
 	amount_per_transfer_from_this = 10
 	list_reagents = list(/datum/reagent/medicine/mutadone = 10)
--- a/fulp_modules/jobs/deputy/code/deputy.dm
+++ b/fulp_modules/jobs/deputy/code/deputy.dm
@@ -178,7 +178,7 @@ GLOBAL_LIST_INIT(available_deputy_depts, sortList(list(SEC_DEPT_ENGINEERING, SEC
 /// Used for Science Deputies
 /obj/item/reagent_containers/hypospray/medipen/mutadone
 	name = "mutadone medipen"
-	desc = "Hulked lings in the RD office? Space adapt traitors bombing the Armory? You know what to do! Comes with 2 uses."
-	volume = 60
-	amount_per_transfer_from_this = 30
-	list_reagents = list(/datum/reagent/medicine/mutadone = 60)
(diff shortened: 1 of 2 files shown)


JohnFulpWillard:
image


In all seriousness, I made it 60u so it would prevent them from getting mutations for a good while afterwards. The sprites for the pens were a big fuckup on my part, but I'd either like them to still be able to get 2 uses out of it, or get double the pens. Mutadone medipens are instant and go through hardsuits, which is why they're unrefillable, which is why I want them to at least get 2 uses out of it a round.

Helianthus00:
For balance reasons, I'd prefer they keep two uses of mutadone, either in one pen or two. And obviously a new sprite would be best but I won't require that for a merge. I like this overall though.

EgorDinamit:
I'd personally make it a 3 use 15u medipen.
5u is more than enough to be able to revert mutations.

fulp_modules/jobs/deputy/code/deputy.dm
```diff
@@ -178,7 +178,10 @@ GLOBAL_LIST_INIT(available_deputy_depts, sortList(list(SEC_DEPT_ENGINEERING, SEC
 /// Used for Science Deputies
 /obj/item/reagent_containers/hypospray/medipen/mutadone
 	name = "mutadone medipen"
-	desc = "Hulked lings in the RD office? Space adapt traitors bombing the Armory? You know what to do! Comes with 2 uses."
```
Helianthus00, line 181: I'd like to retain these two uses. Either through doubling the volume or making a second pen.

ATH1909:
Science deputies now start each shift with two mutadone medipens instead of one.

PR #124, opened 2021-04-15
Deputy update 2

JohnFulpWillard:
## About The Pull Request

1) Gives them access to holding cells back, removes Security Office (Unless on lowpop) and gives them Forensics so they can still access Security records. 
2) They now have their own bowman headsets for their respective departments, which doesn't come with Security comms.
3) Deputies' trims now match their department
4) Some general re-balancing has been done. Deputy's blocking has been increased (90% -> 95%) and Supply Deputies have been given an armor vest and export scanner.
5) Service deputies have been added as Admin/centcom official only, with only 2 lines of code needed to fully enable it.

## Why It's Good For The Game

1) Deputies are powergaming Security gear, which we don't want. I've already made Deputies spawn with all the gear they need, they have no reason to be opening a security locker like Officers do.
2) Deputies have no business to be in Security comms, they're privately working for their assigned department, not the security team. The only exception is for red alert, when they listen to the HoS, which goes for literally every other job because red-alert threats (Nukies, cult, ect) are all things that deputized people can valid anyways.
3) Why not, I've been told Security officers have it, and I thought it would be another way to tell which department a deputy is in, if it isn't already obvious enough. The department means everything to the Deputy.
4) Supply deputies have gotten nothing special while all the others got cool things, thought this fit well for them
5) :)

## Changelog
:cl:
add: Deputies now get a trim that matches their department.
balance: Deputies no longer have Security Office access, but now have Holding Cell access.
balance: Deputies no longer have Security radio comms.
balance: Supply Deputies now has an armor vest.
balance: Deputies' block chance has been slightly increased (90% -> 95%).
/:cl:


Helianthus00:
Considering there is already a pr up for the medipen change, I'd prefer that part be left to the other pr unless it ends up being closed.

PR #126, opened 2021-04-15
Fixes double Sol, removes Bloodsucker Gamemode + other fixes.

JohnFulpWillard:
I've added the commented out 'Vampire Clan' from the code, to use it for dealing with assigning/removing Sol, it seemed to be bugging out due to several roundstart Bloodsuckers, or probably the check that's supposed to look for it just doesn't work at all due to it previously using game_mode to check if it already exists, but the datum/antagonist itself is per the antag, rather than per the gamemode.

## Changelog
:cl:
fix: Sol will no longer randomly cause Bloodsuckers to level up twice.
fix: Bloodsucker masquerade can no longer be abused for infinite flash protection.
fix: Bloodsuckers will no longer lose their antagonist status twice upon their bloodsucker being removed.
fix: Flypeople/Lizard Bloodsuckers will no longer have their lisp removed.
fix: Plasmamen can no longer roll Bloodsucker
del: Bloodsucker and Traitorsucker gamemodes have been removed, as all gamemodes will be removed soon.
del: Bloodsuckers no longer get a bunch of random buffs for literally no reason.
/:cl:

My quest to remove sleep() from Bloodsucker code, continues.


fulp_modules/bloodsuckers/code/antagonists/monsterhunter_datum.dm
```diff
@@ -6,7 +6,7 @@
 #define STATUS_EFFECT_HUNTERPINPOINTER /datum/status_effect/agent_pinpointer/hunter_edition
 
 /datum/antagonist/monsterhunter
-	name = "Monster Munter"
```
EgorDinamit, line 9: Hilarious. That's it.

PR #130, opened 2021-04-19
Doesnt make nightmares dumb unless its an actual nightmare

JohnFulpWillard:
## Why It's Good For The Game

A wizard apparently got stuck at their base or something? I was able to use the things there without the trait, but whatever, they'll need it while on station anyways.

## Changelog
:cl:
fix: Wizard nightmares aren't dumb anymore
/:cl:


fulp_modules/nightmare_edit/nightmare.dm
```diff
@@ -1,4 +1,7 @@
+/// Midround nightmare spawns are stupid.
 /datum/species/shadow/nightmare/spec_life(mob/living/carbon/human/H)
-	. = ..()
+	if(!H.mind.has_antag_datum(/datum/antagonist/nightmare))
```
EgorDinamit, line 3: Make sure nightmare antagonist status is assigned **before** species is.

EgorDinamit (commented):
Well, would you look at that?
image
And actually, you might want to make it properly. If I am not wrong, spec_life is called every tick. You **really** don't want to put such code in there, it should be in gaining race or better yet, in nigthmare **antagonist** code.

fulp_modules/nightmare_edit/nightmare.dm
JohnFulpWillard: Ive tested it and it works.

JohnFulpWillard:
> 
> 
> You might want to make it properly. If I am not wrong, spec_life is called every tick. You **really** don't want to put such code in there, it should be in gaining race or better yet, in nightmare **antagonist** code.

Ive tried making it tied to the antag but nothing I've done actually worked for removing traits. Adding traits work, however.

ATH1909:
> > You might want to make it properly. If I am not wrong, spec_life is called every tick. You **really** don't want to put such code in there, it should be in gaining race or better yet, in nightmare **antagonist** code.
> 
> Ive tried making it tied to the antag but nothing I've done actually worked for removing traits. Adding traits work, however.

did you try using apply_innate_effects()?

PR #141, opened 2021-04-26
Heliostation sm waste and scrubber pipe layer fixes

EIB0N:
## About The Pull Request

This changed the smart pipes for the scrubbers and exhaust from layer 3 to layer 2 like the rest of the scrubbers around the station

## Why It's Good For The Game

Help out the Engineering boys be more organized

## Changelog
:cl:
fix: SM Scrubber and exhaust are now layer 2!
/:cl:


Helianthus00:
image
As far as I know, this does not work. A far better solution is to just change the color to something else that is not used in this area. Then I'd keep the rest of the pipes on layer 3 as normal so that the rest of the layers are reserved for extra pipes.

Helianthus00 (changes requested):
image
As far as I know, this doesn't work. You need to change the color of the pipes in the wastenet instead and leave everything else on layer 3.

EIB0N:
Fixed it

Helianthus00 (changes requested):
image
This gas meter is on the wrong layer. Please make sure you test this as well to make sure there aren't unintended issues. I still don't like the use of layer 2 scrubber pipes in the chamber.

EIB0N (commented):
gas flow meter changed to layer 2

Helianthus00 (changes requested):
image
Making this pipe brown means it will no longer connect with the red pipes. You need to change it back to red or it won't work.

tgstation.dme
```diff
@@ -3637,7 +3637,6 @@
 #include "fulp_modules\maps\cytology_spawner\beefcyto.dm"
 #include "fulp_modules\maps\fulp_shuttles\fulp_shuttles.dm"
 #include "fulp_modules\ZFulpcode\modules\jobs\huds.dm"
-#include "fulp_modules\ZFulpcode\modules\jobs\inventories.dm"
```
Helianthus00, line 3640: This shouldn't be removed.

Helianthus00 (changes requested):
You don't need pipe bridges here. They are only needed if the pipes are the same layer.

tgstation.dme
EIB0N: readded it

EIB0N (commented):
Fixed the pipe colors and removed the bridges

tgstation.dme
Helianthus00: it needs to be in alphabetical order

Helianthus00 (approved):

PR #143, opened 2021-04-29
Inspecting Weapons Shows Stats

SpaceDragon00:
## About The Pull Request

Modifies the item examine proc to show an offensive tag on weapons. This works like an armor tag, and details the stats of the weapon in-game for easy reference. 

## Why It's Good For The Game

It can be hard for beginners to differentiate a lot of the melee weapons. There's no real easy way to test out damage in-game, and the TG wiki is often outdated, if it includes the weapon at all. This PR attempts to fix this by simply giving the player relevant data on any weapon they have.

## Changelog
:cl:
add: Added a weapon stats tab when examining weapons
/:cl:


tgstation.dme
```diff
@@ -3639,6 +3639,7 @@
 #include "fulp_modules\maps\fulp_shuttles\fulp_shuttles.dm"
 #include "fulp_modules\ZFulpcode\modules\jobs\huds.dm"
 #include "fulp_modules\ZFulpcode\modules\jobs\inventories.dm"
+#include "fulp_modules\edits\item_desc\item_desc_edit.dm"
```
SgtHunk, line 3642: .dme files are ordered alphabetically, this will make it so every following PR moves this file to where it's supposed to be due to it being re-ordered automatically

SgtHunk (changes requested):

tgstation.dme
SpaceDragon00: Simply updated the PR with the auto organized dme file, moving it to the correct place, I hope

Helianthus00:
Wouldn't this make more sense to PR upstream? I like this change a lot, but unless they have explicitly denied it, I'd rather it be an upstream thing.

SpaceDragon00:
> 
> 
> Wouldn't this make more sense to PR upstream? I like this change a lot, but unless they have explicitly denied it, I'd rather it be an upstream thing.

I'll change it to PR it upstream, then, and leave this open in case it's denied. The main reason I didn't is because I felt like the information (or lack thereof) provided by TG was fairly deliberate, especially with that horribly esoteric 'Force' system you get when you hover over something

SpaceDragon00:
https://github.com/tgstation/tgstation/pull/58845

Code improved and PRed to TG, let's see how it goes

SpaceDragon00:
Forgot about this to be honest. TG seems amenable to the feature with some changes, expect it upstream when I become less of a lazy piece of garbage

PR #148, opened 2021-05-04
Fulp logo

JohnFulpWillard:
## About The Pull Request

Ports the Fulpstation logo from the old repo, this time, I hope, better.

## The problem with this Pull Request

This moves ALL plaques to use our .dmi file, because I couldnt figure out how to get it to work otherwise, they were loading the icon_state in the map, rather than in their own file.
Currently, only TWO plaques exist;
- Space Station 13 logo on every map (Which is what this PR changes)
- Derilict (Russian) Drone station's logo

These don't cause Runtimes if missing, they just don't appear at all, but I added them to our .dmi file so they would anyways. 
The ACTUAL problem, is if TG now also adds more plaques, THEN they'll be invisible to us, but as said earlier, no runtimes or errors, just... not there.


ATH1909:
https://www.youtube.com/watch?v=_J6-3l3hCm0

PR #150, opened 2021-05-04
[SeleneStation] Swaps Genetics - Toxins storage, and 2 other things.

JohnFulpWillard:
## About The Pull Request

Swaps Genetics with Toxins storage, adding windows that the main halls can see.
Additionally, adds Sunglasses to the Toxins range area, to match with Heliostation, replaces the ancient heirloom toolbox in Engineering maint, with a regular non-heirloom toolbox, and fixes the disposal unit in the Nanite lab not being connected to disposals.

image
image

## Why It's Good For The Game

Probably isn't, I'm just biased and like it when people can ask me for genes, which they rarely do when they can never see the area I'm at.

## Changelog
:cl:
add: Sunglasses has been added to Toxins range on Selene.
qol: Genetics and Toxins storage areas have been swapped on Selene.
fix: The ancient toolbox in Engineering maints is no longer an heirloom.
/:cl:

### DIFF
--- a/_maps/map_files/SeleneStation/SeleneStation.dmm
+++ b/_maps/map_files/SeleneStation/SeleneStation.dmm
@@ -1335,39 +1335,6 @@
 "asf" = (
 /turf/open/floor/iron/dark,
 /area/maintenance/port/aft)
-"asi" = (
-/obj/effect/turf_decal/tile/purple{
-	color = "#4D0067";
-	dir = 4;
-	name = "Science purple corner"
-	},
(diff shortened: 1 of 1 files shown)


Helianthus00:
I'm not opposed to this change but also I think the new rooms are less interesting and have boring shape. If you remake the rooms instead of just swapping them, I'd like this more.

Helianthus00:
I don't like this. If you re-do it by remaking the rooms and changing the structure of the department, I will consider that, but this doesn't work for me.

PR #156, opened 2021-05-10
[Testmerge only] Revolutionaries no longer turn into Enemies of the Revolution (Not state)

JohnFulpWillard:
## About The Pull Request

Meant for testmerge only, to know if https://github.com/tgstation/tgstation/pull/59015 is working as intended.


JohnFulpWillard:
tested and worked.

PR #163, opened 2021-05-13
Bloodsucker fixes (sleep is too strong)

### LINKED ISSUE #162: Bloodsuckers get stuck at rank 7
Round ID: [14531](https://scrubby.melonmesa.com/round/14531)
Testmerges:

TGU - Gamemode removal, GAGS & More #159
Reproduction:

Get to level 7
Close a coffin
Rank up to level 7 again
Enter Sol
Rank up to level 7 again
It's level 7 Bloodsucker hell.

If anyone can fix this, be my guest. I haven't looked into it yet but I'm putting this issue up so I don't forget about it.

I somehow posted this on the wrong fucking codebase. I have too many tabs open for this.

JohnFulpWillard:
## About The Pull Request

I hate sleep but I also dont know any other alternative I could use here.

Fixes #162 so thank you for opening that issue to bring this to my attention!

## Why It's Good For The Game

When the Antagonist is Playable ! 😳 

## Changelog
:cl:
fix: Bloodsucker passive abilities will no longer cancel if they get shove/stunned, will no longer drain immense amounts of Blood randomly, and will no longer prevent construction
fix: Bloodsuckers will no longer be able to infinitely buff themselves once they reach Rank 7
balance: Bloodsuckers can now remove their wounds by going through Torpor
fix: Recuperate now should work properly again.
/:cl:

### DIFF
--- a/fulp_modules/main_features/bloodsuckers/code/antagonists/bloodsucker_datum.dm
+++ b/fulp_modules/main_features/bloodsuckers/code/antagonists/bloodsucker_datum.dm
@@ -294,6 +294,7 @@
 
 /datum/antagonist/bloodsucker/proc/SpendRank()
 	set waitfor = FALSE
+
 	if(bloodsucker_level_unspent <= 0 || !owner || !owner.current || !owner.current.client)
 		return
 	//TODO: Make this into a radial, or perhaps a tgui next UI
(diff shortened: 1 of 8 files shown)


EgorDinamit:
### "Athath says:"

I think you should use callbacks instead of sleep(), since it may cause some unforeseen consequences.

JohnFulpWillard:
> 
> ### "Athath says:"
> 
> I think you should use callbacks instead of sleep(), since it may cause some unforeseen consequences.

One day. for now, I just want it to be playable.

PR #176, opened 2021-05-18
Tom is now more powerful than ever

JohnFulpWillard:
## About The Pull Request

AS MY FIRST ACT OF BEING THE NEW OWNER OF FULPSTATION;

I love Tom, what would make them even better is if you can play with them, otherwise they just stand around and give permas a mood boost.

## Changelog
:cl:
add: Tom has a new AI!
/:cl:


Helianthus00:
I really would prefer something like this be pr'd to TG instead.

JohnFulpWillard:
I'll hopefully get around to asking TG one day, but adding it there would require a ton of changes to the AI in general, which is why I'm hesitating currently.

PR #191, opened 2021-05-31
[NON-MODULAR] Felinids are now as fragile as a cat

Kitsunemitsu:
## About The Pull Request
Felinids are now Fragile. Cats can't eat fruit or vegetables, so I made them toxic to the cat.
These are both one line edits, so they can be added or removed at your pleasure

## Why It's Good For The Game
This way, people will have disincentives to be playing an ERP race.

## Changelog
:cl:
balance: Made Felinids take 3x Brute
balance: Fruit and Vegetables are now toxic to cats
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/felinid.dm
+++ b/code/modules/mob/living/carbon/human/species_types/felinid.dm
@@ -12,9 +12,10 @@
 	changesource_flags = MIRROR_BADMIN | WABBAJACK | MIRROR_PRIDE | MIRROR_MAGIC | RACE_SWAP | ERT_SPAWN | SLIME_EXTRACT
 	species_language_holder = /datum/language_holder/felinid
 	disliked_food = GROSS | RAW | CLOTH
+	toxic_food = FRUIT | VEGETABLES //Fulpstation Felinid Edit
+	brutemod = 3 //Fulpstation Felinid Edit
 	var/original_felinid = TRUE //set to false for felinids created by mass-purrbation
 	payday_modifier = 0.75
-	brutemod = 3 //Fulpstation Felinid Edit
 	ass_image = 'icons/ass/asscat.png'
 	family_heirlooms = list(/obj/item/toy/cattoy)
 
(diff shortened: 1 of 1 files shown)


SgtHunk:
As hilarious as toolboxing a felinid to crit in 3 hits is, no. This is both
a) unbalanced, and discourages people whilst having no benefits unlike plasmamen's weakness.
b) non-modular, when it could be made with our ways of overwriting variables.

It'd be better to disable the species tbh, but that requires coordination between config-access admins and contributors, which blue names can take care of.

StoneMason:
Whilst I'd love felinid changes, I have some gripes.
1. Originally, the cocoa toxic reaction in felinids led to Reddit brigading TG servers to force feed felinids chocolate, so it was reverted. It would be better to have them not like the food, rather than have it be toxic, to avoid a situation like this. 
2. How would this interact with having the fragile quirk? 
3. 3x Brute for a race with no significant advantages? It'd take 1-2 dsword attacks and they're cleaned out. Other races with brute modifiers have advantages like revival or no sickness/hunger.

Kitsunemitsu:
I'm reducing from 3x to 1.25x.
We really should disable them though

Kitsunemitsu:
It wasn't supposed to be a "balance" it was supposed to push people away from playing felinid.

SgtHunk:
> It wasn't supposed to be a "balance" it was supposed to push people away from playing felinid.

Why? As far as I know, you don't play here anymore.
Furthermore, not even you push away people from playing the species on your server.
Considering you gave them a speed boost in exchange for 1.2x brute modifier.
If we wanted to disable and stop people from playing a species, we'd disable and TM a removal to reset preferences.

Kitsunemitsu:
I was just reading some things, Pucci.

Willard actually was talking about removing felinids.
Personally, I think that this would stop people from playing something you always talk about removing without actually causing people who play felinids to get too mad.
Eventually, no one will play felinid, and no one would care about it being removed.

SgtHunk:
> If we wanted to disable and stop people from playing a species, we'd disable and TM a removal to reset preferences.

Sheits:
> I'm reducing from 3x to 1.25x.
> We really should disable them though

"we"

Kitsunemitsu:
Alright, I'll close this

PR #194, opened 2021-05-31
Changes how Torpor is dealt with

JohnFulpWillard:
## About The Pull Request

How Torpor works now
---

Torpor is triggered by:

- Being in a Coffin while Sol is on
- Entering a Coffin with more than 10 combined Brute/Burn damage
- Dying

Torpor is ended by:

- Having less than 10 Brute damage while OUTSIDE of your Coffin while it isnt Sol
- Having less than 10 Brute & Burn Combined while INSIDE of your Coffin while it isnt Sol
- Sol being over

Also, makes Torpor give you immunity to low pressure, so while flamethrowers/fire will still be an annoyance, you wont take damage to, for example, space. I've had 3 people complain already about being endlessly stuck in Torpor because they thought it would be a good idea to enter Torpor while in a depressurizing room.

### DIFF
--- a/fulp_modules/main_features/bloodsuckers/code/bloodsucker/bloodsucker_life.dm
+++ b/fulp_modules/main_features/bloodsuckers/code/bloodsucker/bloodsucker_life.dm
@@ -347,14 +347,13 @@
 				to_chat(owner.current, "<span class='userdanger'>You are staked! Remove the offending weapon from your heart before sleeping.</span>")
 				return
 			/// Otherwise, check if it's Sol, to enter Torpor.
-			if(clan.bloodsucker_sunlight.amDay && !HAS_TRAIT(owner.current, TRAIT_NODEATH))
+			if(clan.bloodsucker_sunlight.amDay)
 				Check_Begin_Torpor(TRUE)
 		/// You are in Torpor, and in a Coffin. Check if it's not Daytime & you have less than 10 Brute/Burn combined to end Torpor.
 		else if(!clan.bloodsucker_sunlight.amDay && total_damage <= 10)
 			Check_End_Torpor()
-			return
 	/// You're not in a Coffin, but are in Torpor. Check if it's not Daytime, & you have less than 10 Brute (NOT Burn) to end Torpor.
-	if(!clan.bloodsucker_sunlight.amDay && total_brute <= 10 && HAS_TRAIT(owner.current, TRAIT_NODEATH))
+	else if(!clan.bloodsucker_sunlight.amDay && total_brute <= 10 && HAS_TRAIT(owner.current, TRAIT_NODEATH))
 		Check_End_Torpor()
 
 /datum/antagonist/bloodsucker/proc/Check_Begin_Torpor(CheckChecks = FALSE)
--- a/fulp_modules/main_features/bloodsuckers/code/bloodsucker/bloodsucker_life.dm
+++ b/fulp_modules/main_features/bloodsuckers/code/bloodsucker/bloodsucker_life.dm
@@ -324,7 +324,7 @@
 
 	Torpor is triggered by:
 	- Being in a Coffin while Sol is on, dealt with by /HandleTorpor()
-	- Entering a Coffin with more than 10 combined Brute/Burn damage, dealt with by /closet/crate/coffin/close() [bloodsucker_coffin.dm] - This will give a Radial Yes/No menu on whether they want to enter Torpor.
+	- Entering a Coffin with more than 10 combined Brute/Burn damage, dealt with by /closet/crate/coffin/close() [bloodsucker_coffin.dm]
 	- Death, dealt with by /HandleDeath()
 	Torpor is ended by:
 	- Having less than 10 Brute damage while OUTSIDE of your Coffin while it isnt Sol, dealt with by /HandleTorpor()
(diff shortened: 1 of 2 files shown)


fulp_modules/main_features/bloodsuckers/code/bloodsucker/bloodsucker_life.dm
```diff
@@ -330,27 +346,29 @@
...
 				return
-			/// Otherwise, check for Sol, or if injured enough to enter Torpor.
-			if(clan.bloodsucker_sunlight.amDay || total_damage >= 10)
-				to_chat(owner.current, "<span class='notice'>You enter the horrible slumber of deathless Torpor. You will heal until you are renewed.</span>")
-				Torpor_Begin()
-	/// If it's not Sol and you have 0 Brute damage, check to End Torpor.
-	if(!clan.bloodsucker_sunlight.amDay && total_brute <= 10)
-		if(HAS_TRAIT(owner.current, TRAIT_NODEATH))
+			/// Otherwise, check if it's Sol, to enter Torpor.
+			if(clan.bloodsucker_sunlight.amDay && !HAS_TRAIT(owner.current, TRAIT_NODEATH))
```
SpaceDragon00, line 350: You've already checked if you don't have NODEATH on line 344

fulp_modules/main_features/bloodsuckers/code/bloodsucker/bloodsucker_life.dm
```diff
@@ -330,27 +346,29 @@
...
+				Check_Begin_Torpor(TRUE)
+		/// You are in Torpor, and in a Coffin. Check if it's not Daytime & you have less than 10 Brute/Burn combined to end Torpor.
+		else if(!clan.bloodsucker_sunlight.amDay && total_damage <= 10)
```
SpaceDragon00, line 353: 353-355 is unnecessary. The conditions to check the end of torpor are the same regardless of whether or not you are in a coffin, so you don't need separate blocks for whether or not you're in one

JohnFulpWillard: The check below what you want to remove in favor of, is only taking brute damage into account. If you're in a coffin with 0 brute and 15 burn damage, you shouldn't be exiting Torpor.

fulp_modules/main_features/bloodsuckers/code/bloodsucker/bloodsucker_life.dm
```diff
@@ -330,27 +346,29 @@
...
-			/// Otherwise, check for Sol, or if injured enough to enter Torpor.
-			if(clan.bloodsucker_sunlight.amDay || total_damage >= 10)
-				to_chat(owner.current, "<span class='notice'>You enter the horrible slumber of deathless Torpor. You will heal until you are renewed.</span>")
-				Torpor_Begin()
-	/// If it's not Sol and you have 0 Brute damage, check to End Torpor.
-	if(!clan.bloodsucker_sunlight.amDay && total_brute <= 10)
-		if(HAS_TRAIT(owner.current, TRAIT_NODEATH))
+			/// Otherwise, check if it's Sol, to enter Torpor.
+			if(clan.bloodsucker_sunlight.amDay && !HAS_TRAIT(owner.current, TRAIT_NODEATH))
+				Check_Begin_Torpor(TRUE)
```
SpaceDragon00, line 351: Check_Begin_Torpor on 351 here will only fire if it's daytime. It will then fail because it'll only pass if it's NOT daytime

Set this to false and add the damage as an additiona conditional to the if statement on 350

fulp_modules/main_features/bloodsuckers/code/bloodsucker/bloodsucker_life.dm
```diff
@@ -330,27 +346,29 @@
...
-	/// Disable ALL Powers
-	for(var/datum/action/bloodsucker/power in powers)
-		if(power.active && !power.can_use_in_torpor)
-			power.DeactivatePower()
+			return
+	/// You're not in a Coffin, but are in Torpor. Check if it's not Daytime, & you have less than 10 Brute (NOT Burn) to end Torpor.
+	if(!clan.bloodsucker_sunlight.amDay && total_brute <= 10 && HAS_TRAIT(owner.current, TRAIT_NODEATH))
+		Check_End_Torpor()
+
+/datum/antagonist/bloodsucker/proc/Check_Begin_Torpor(CheckChecks = FALSE)
```
SpaceDragon00, line 359: It'd probably be better to call CheckChecks something like SkipChecks or Force, cause only Checking the Checks when "CheckChecks" is false can be a bit confusing

fulp_modules/main_features/bloodsuckers/code/bloodsucker/bloodsucker_life.dm
```diff
@@ -330,27 +346,29 @@
...
+	var/mob/living/carbon/user = owner.current
+	var/total_brute = user.getBruteLoss_nonProsthetic()
+	var/total_burn = user.getFireLoss_nonProsthetic()
+	var/total_damage = total_brute + total_burn
+	/// Checks - Not daylight & Has more than 10 Brute/Burn & not already in Torpor
+	if(!clan.bloodsucker_sunlight.amDay && total_damage >= 10 && !HAS_TRAIT(owner.current, TRAIT_NODEATH))
+		Torpor_Begin()
 
 /datum/antagonist/bloodsucker/proc/Check_End_Torpor()
 	/// You're not in Sol? (Slept in a Locker for example), then you don't need to leave it.
```
SpaceDragon00, line 373: "You're not in Sol?" -> "You're not in torpor?"

JohnFulpWillard: No, you're not in Sol. You can sleep through Sol in a locker.

fulp_modules/main_features/bloodsuckers/code/bloodsucker/bloodsucker_life.dm
```diff
@@ -330,27 +346,29 @@
...
-	ADD_TRAIT(owner.current, TRAIT_FAKEDEATH, BLOODSUCKER_TRAIT)
-	ADD_TRAIT(owner.current, TRAIT_DEATHCOMA, BLOODSUCKER_TRAIT)
-	owner.current.Jitter(0)
-	/// Disable ALL Powers
-	for(var/datum/action/bloodsucker/power in powers)
-		if(power.active && !power.can_use_in_torpor)
-			power.DeactivatePower()
+			return
+	/// You're not in a Coffin, but are in Torpor. Check if it's not Daytime, & you have less than 10 Brute (NOT Burn) to end Torpor.
+	if(!clan.bloodsucker_sunlight.amDay && total_brute <= 10 && HAS_TRAIT(owner.current, TRAIT_NODEATH))
```
SpaceDragon00: Make this an else if, since it's only supposed to run when you're not in a coffin

fulp_modules/main_features/bloodsuckers/code/bloodsucker/bloodsucker_life.dm
SpaceDragon00: Comment was made before the changes, oopsies

SpaceDragon00: Or wait no, this was referring to line 374, which is collasped by default. The comment "You're not in sol" on 373 is above an if statement that checks for NODEATH on line 374

fulp_modules/main_features/bloodsuckers/code/bloodsucker/bloodsucker_life.dm
```diff
@@ -330,27 +346,28 @@
...
+		return
+	var/mob/living/carbon/user = owner.current
+	var/total_brute = user.getBruteLoss_nonProsthetic()
+	var/total_burn = user.getFireLoss_nonProsthetic()
+	var/total_damage = total_brute + total_burn
+	/// Checks - Not daylight & Has more than 10 Brute/Burn & not already in Torpor
+	if(!clan.bloodsucker_sunlight.amDay && total_damage >= 10 && !HAS_TRAIT(owner.current, TRAIT_NODEATH))
+		Torpor_Begin()
 
 /datum/antagonist/bloodsucker/proc/Check_End_Torpor()
```
SpaceDragon00, line 372: Check_End_Torpor should probably have SkipChecks like Check_Begin_Torpor since it is often called when the variables that end torpor are already evaluated, like on 357 and 354

Note that Check_End_Torpor() will end torpor whenever it's called if you're not in a coffin, since 377 does not check for brute damage as of the time of review

This will still function correctly since Check_End_Torpor is only called outside of a coffin in HandleTorpor when you have less than 10 brute, but it'd be good to have the checks standardized

JohnFulpWillard: Check_End_Torpor cancels out if you're not in Torpor already.
if(!HAS_TRAIT(owner.current, TRAIT_NODEATH))
		return

remove32:
john are we good on this one? it works?

JohnFulpWillard:
> john are we good on this one? it works?

Yep, all good

PR #198, opened 2021-06-01
SeleneStation: Adding air alarm to HFR

EIB0N:
## About The Pull Request

Adding an air alarm to HFR

## Why It's Good For The Game

Its a nice addition also standards

## Changelog
:cl:
add: HFR room now has an air alarm

/:cl:


Helianthus00:
You should just change the area to be the atmos engine area. That would make things a bit more clear anyway. Otherwise, there will be two air alarms for that one area, which can get confusing.

EIB0N:
> You should just change the area to be the atmos engine area. That would make things a bit more clear anyway. Otherwise, there will be two air alarms for that one area, which can get confusing.

done

PR #206, opened 2021-06-05
Monster hunters spawn, get buffed Fortitude, and no longer take blood to track monsters

JohnFulpWillard:
## About The Pull Request

dies

## Changelog
:cl:
fix: Monster hunters now spawn!
balance: Flow now gives damage resistance
balance: Monster Hunters no longer use Blood to track Monsters
/:cl:

### DIFF
--- a/fulp_modules/main_features/bloodsuckers/code/gamemodes/monsterhunter_gamemode.dm
+++ b/fulp_modules/main_features/bloodsuckers/code/gamemodes/monsterhunter_gamemode.dm
@@ -45,7 +45,7 @@
 		break
 
 /datum/round_event/monster_hunters/proc/check_for_monsters()
-	for(var/mob/players in GLOB.alive_mob_list)
+	for(var/mob/living/carbon/human/players in GLOB.alive_mob_list)
 		/// Make sure there are monsters on the station, otherwise don't spawn them in
 		if(IS_CULTIST(players) || IS_HERETIC(players) || IS_BLOODSUCKER(players) || IS_WIZARD(players) || players.mind.has_antag_datum(/datum/antagonist/changeling))
 			message_admins("MONSTERHUNTER NOTICE: A Monster Hunter is attempting to awaken.")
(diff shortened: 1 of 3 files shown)


fulp_modules/main_features/bloodsuckers/code/gamemodes/monsterhunter_gamemode.dm
```diff
@@ -23,7 +23,7 @@
 	fakeable = FALSE
 
 /datum/round_event/monster_hunters/start()
-	for(var/mob/living/carbon/human/H in shuffle(GLOB.player_list))
+	for(var/mob/living/H in GLOB.player_list)
```
EgorDinamit, line 26: This makes it so simple mobs can become antagonist.

The entire code is kinda wacky as well, you'd want to fix it.

fulp_modules/main_features/bloodsuckers/code/gamemodes/monsterhunter_gamemode.dm
```diff
@@ -23,7 +23,7 @@
...
 		/// Make sure there are monsters on the station, otherwise don't spawn them in.
 		if(!H.mind.has_antag_datum(/datum/antagonist/bloodsucker) || !H.mind.has_antag_datum(/datum/antagonist/changeling) || !H.mind.has_antag_datum(/datum/antagonist/heretic) || !H.mind.has_antag_datum(/datum/antagonist/cult) || !H.mind.has_antag_datum(/datum/antagonist/wizard))
```
EgorDinamit, line 28: You are checking if current mob in the loop doesn't have these antag datums.
Know what's the issue here? You are checking it like this:
"If NOT bloodsucker OR NOT changeling OR NOT cultist".

You see the issue yet? It will always return true because if, for example, the mob is a bloodsucker it will then check if it's a changeling. Because you are using OR - it will always break the loop.
Another issue here is with the break itself, you are stopping the loop if some random mob in the list doesn't have antagonist status and prevents the loop from checking the other mobs.

So, you really want to fix it if you expect it to work.

PR #215, opened 2021-06-08
[Selene] Edits (buffs) Chef's CQC

JohnFulpWillard:
## About The Pull Request

In black - old version
In magenta - new version
image

yes it works in Botany now.

## Why It's Good For The Game

I dont like Bartenders being shot in their own bar, or CQC not working in the bar's atrium because glass panes are meant to show that this is actually the theatre, but I'm also biased against Botanists because I just hate them all for making nothing but deathnettle.

Also there was no bar atrium or kitchen coldroom (despite that being an additional CQC area (why is it an additional area its done automatically)) so those have been added.

## Changelog
:cl:
balance: The chef's CQC in Selene works in the bar's full atrium and botany, but no longer in the bar itself.
balance: The table between the Kitchen and Botany has been replaced with a door.
/:cl:

### DIFF
--- a/_maps/map_files/SeleneStation/SeleneStation.dmm
+++ b/_maps/map_files/SeleneStation/SeleneStation.dmm
@@ -824,7 +824,7 @@
 "alJ" = (
 /obj/machinery/holopad,
 /turf/open/floor/iron/showroomfloor,
-/area/service/kitchen)
+/area/service/kitchen/coldroom)
 "alL" = (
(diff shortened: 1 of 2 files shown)


Helianthus00 (changes requested):
You'll need to add an APC for the bar atrium since it's a new area, as well as the kitchen coldroom.

remove32:
i take exception with all botanists being called shithead deathnettlers. when i play botany the first step is always to stock the fridge with a variety and quantity of items. THEN the death nettles come out.

PR #218, opened 2021-06-11
Beefmen can no longer teleport to NOTELEPORT areas

SgtHunk:
## About The Pull Request

Adds another check to the proc that returns the turf that the beefman phobetor tear can choose. Also documents it and makes it a bit more readable.
tl;dr beefmen can't teleport to areas that have the NOTELEPORT flag (CentCom, wizard's den, etc)

## Why It's Good For The Game

Finally got the motivation to do this
Also futureproofing in case that the non-human nuke ops PR at TG gets merged. (ALSO YES THIS FIXES WIZARD BEEFMAN, DONT REMIND ME OF WIZARD BREAKING INTO CENTCOM)

## Changelog
:cl:
fix: Beefmen can no longer teleport to areas that they shouldn't be able to teleport to.
/:cl:

### DIFF
--- a/fulp_modules/main_features/beefmen/code/modules/mob/beefman_trauma.dm
+++ b/fulp_modules/main_features/beefmen/code/modules/mob/beefman_trauma.dm
@@ -25,16 +25,29 @@
 		return target
 	return null
 
-/proc/check_turf_is_valid(turf/T, checkFloor = TRUE)
+/**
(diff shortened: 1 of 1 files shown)


remove32:
NOOO MY BROKEN BEEFY BOYS

good change. i like that you updated the var names to be more descriptive. if we had GBP i'd give you some

PR #229, opened 2021-06-16
fulp_config hardcoding

JohnFulpWillard:
## About The Pull Request

???


remove32:
if you're intent on keeping headofpseudostaff we need to know how that will show in the config files otherwise it'll be an unused variable with no value

JohnFulpWillard:
> 
> 
> if you're intent on keeping headofpseudostaff we need to know how that will show in the config files otherwise it'll be an unused variable with no value

idk (i dont have config access)

remove32:
you should in the test server. even without having access to config in prod you should be able to know how the .DM files will interact with the config. 
By convention, .DMs will never push data into config files.

PR #245, opened 2021-06-26
[Heliostation] Fixes atmos from being sabotaged roundstart

JohnFulpWillard:
## About The Pull Request

These pipes used to be blue, so it would connect to the pipe UNDER the gas pump, rather than to the gas pump.
image

## Why It's Good For The Game

🚎


Helianthus00 (approved):
fucking smartpipes

PR #246, opened 2021-06-26
Fixes admins giving Bloodsucker objectives & Adds monster hunters to admin list

JohnFulpWillard:
## About The Pull Request

Admins giving Bloodsucker objectives will now no longer say 'Nothing', and they can now give the Monster hunter's objective to destroy all monsters.

### DIFF
--- a/fulp_modules/main_features/bloodsuckers/code/bloodsucker/bloodsucker_objectives.dm
+++ b/fulp_modules/main_features/bloodsuckers/code/bloodsucker/bloodsucker_objectives.dm
@@ -8,10 +8,6 @@
 /datum/objective/bloodsucker
 	martyr_compatible = TRUE
 
-/// GENERATE!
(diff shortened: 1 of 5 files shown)


fulp_modules/main_features/bloodsuckers/code/bloodsucker/bloodsucker_objectives.dm
```diff
@@ -37,6 +33,10 @@
 /datum/objective/bloodsucker/lair
 	name = "claimlair"
 
+/datum/objective/bloodsucker/lair/New()
+	update_explanation_text()
```
SpaceDragon00, line 37: Is there any reason why you cant add this to /datum/objective/bloodsucker/New() instead of adding it to all objectives individually?

JohnFulpWillard: get real

PR #266, opened 2021-07-09
Adds PubbyStation

JohnFulpWillard:
## About The Pull Request

I literally havent changed anything i copy pasted old pubby into strongdmm, saved and hit create pull request.

Don't look at the commits because it's not real

I'll "fulpify" the map in a future PR, this version of it will be treated as if it's on TG.


_maps/pubbystation.json
```diff
@@ -0,0 +1,17 @@
+{
+	"version": 1,
+	"map_name": "Heliostation",
+	"map_path": "map_files/Heliostation",
+	"map_file": "Heliostation.dmm",
```
SpaceDragon00, line 5: It's about time we added Heliostation

JohnFulpWillard: rot in hell

PR #270, opened 2021-07-12
nerfs helio's perma

JohnFulpWillard:
## About The Pull Request

- Removes the telepad (adds another portable flasher to the armory - one was removed in the previous PR, so this is readding)
- Moves sinks from the cells to the bathroom
- Removes 3 bookcases (no longer enough to make a baseball bat)
- Replaces all windows with reinforced windows (can no longer easily break them to make spears with the glass shards)
- Moves space heater to the perma maint area
- Removes the drink vendor (no more greybull)

## Why It's Good For The Game

Fuck perma mains.


ATH1909:
"Removes 3 bookcases (no longer enough to make a baseball bat)"

cringe

PR #271, opened 2021-07-13
Makes antag tips have an action trigger

JohnFulpWillard:
## About The Pull Request

This was a disaster. I tried to make it TGUI but I don't think that's a thing, modularly.


JohnFulpWillard:
Unless I can figure out how to get a verb working to remove it mid-round, I don't think this is a good idea.

PR #285, opened 2021-07-20
Deputy Update 4: Full-departmental Deputies & Secbelt

JohnFulpWillard:
## About The Pull Request

Deputies keep telling me that they are part of security, hopefully this shuts them up finally.

Gives them a security belt with all their gear so they don't have to steal one from a security officer's locker (no more excuses for opening one now).

## Why It's Good For The Game

This is how I intended deputies to be in the first place.

### DIFF
--- a/fulp_modules/main_features/jobs/deputy/code/deputy.dm
+++ b/fulp_modules/main_features/jobs/deputy/code/deputy.dm
@@ -1,11 +1,11 @@
 /datum/job/fulp/deputy
 	title = "Deputy"
 	auto_deadmin_role_flags = DEADMIN_POSITION_SECURITY
-	department_head = list("Head of Security") // Sadly.
+	department_head = list("Head of Security")
 	faction = "Station"
-	total_positions = 4
-	spawn_positions = 4
-	supervisors = "the head of your assigned department, and the head of security, but only on red alert"
+	total_positions = 0
+	spawn_positions = 0
+	supervisors = "the head of your assigned department"
 	selection_color = "#ffeeee"
 	minimal_player_age = 14
 	exp_requirements = 300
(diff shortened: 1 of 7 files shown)


fulp_modules/main_features/jobs/deputy/code/deputy.dm
```diff
@@ -86,118 +147,41 @@
...
-/// Engineering
-/datum/job/fulp/deputy/proc/announce_engineering(mob/deputy, department)
-	var/obj/machinery/announcement_system/announcement_system = pick(GLOB.announcement_systems)
-	if(isnull(announcement_system))
-		return
-	announcement_system.announce_engineering(deputy, department)
+	var/assigned_department = SEC_DEPT_NONE
+	var/channel
 
-/obj/machinery/announcement_system/proc/announce_engineering(mob/deputy, department)
```
JohnFulpWillard, line 152: how did i get away with this

fulp_modules/main_features/jobs/deputy/code/deputy_clothing.dm
```diff
@@ -4,56 +4,44 @@
...
 	shoes = /obj/item/clothing/shoes/laceup
 	l_pocket = /obj/item/flashlight/seclite
-	r_pocket = /obj/item/assembly/flash/handheld
+	r_pocket = /obj/item/pda/security
 	backpack = /obj/item/storage/backpack/security
 	satchel = /obj/item/storage/backpack/satchel/sec
 	duffelbag = /obj/item/storage/backpack/duffelbag/sec
 
+	backpack_contents = list(
+		/obj/item/melee/baton/loaded = 1,
```
ATH1909, line 19: I'd suggest moving the holosign creator here and putting the baton in their secbelt.

... I'd also suggest just giving them a normal secbelt instead of giving them one that has a near-useless (in 99% of circumstances) smoke grenade instead of a flashbang, but that's just me.

ATH1909: also, giving deputies a spare set of handcuffs or two would also probably help suppress the urge to go to sec.

JohnFulpWillard: I had it in the bag instead because security officers had it there, but I'll swap it.

PR #286, opened 2021-07-20
Removes deputies

JohnFulpWillard:
## About The Pull Request

Alternative PR to #285

TG's accesses are fucked and people will abuse it basically no matter what you try to do. Unless rules on what deputies are allowed to take actually get enforced, deputies will not fit here.


Helianthus00:
I'd really prefer to not remove deputies, unless you can no longer find ways around tg's changing of things.

SpaceDragon00:
Im confused about what problem you are specifically having with deputies. I'd like it if you could lay out the specific issue(s) in detail so alternatives can be discussed

JohnFulpWillard:
> 
> 
> Im confused about what problem you are specifically having with deputies. I'd like it if you could lay out the specific issue(s) in detail so alternatives can be discussed

People go to brig, take sec radio and sec gear, then patrol the halls.

Helianthus00:
Merged #285 so I'm closing this

ReinaCoder:
please merge

PR #289, opened 2021-07-22
Bloodsucker fixes

JohnFulpWillard:
## About The Pull Request

I got reports that Recuperate doesn't stop costing blood when deactivated?
If two bloodsuckers try to mesmerize, shit gets fucked up, so I won't allow 2 mesmerizes to go through.
Distress was screeching to all bloodsuckers rather than just their master


JohnFulpWillard:
needs testmerging first

PR #292, opened 2021-07-22
Return of the Banned Beefmen

JohnFulpWillard:
## About The Pull Request

That wasn't a 'meme'.

### DIFF
--- a/fulp_modules/Z_edits/race_bans/races.dm
+++ b/fulp_modules/Z_edits/race_bans/races.dm
@@ -7,7 +7,10 @@
 /// Made into an individual proc to ensure that the to_chat message would always show to users. Sometimes it would not appear during roundstart as it would be sent too soon.
 /mob/living/carbon/proc/banned_species_revert()
 	to_chat(src, span_alert("You are currently banned from playing this race. Please review any ban messages you have received, and contact admins if you believe this is a mistake."))
-	set_species(/datum/species/human)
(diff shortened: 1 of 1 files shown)


SpaceDragon00:
Tom replied to the previous PR with his thoughts on the matter

image

JohnFulpWillard:
> 
> 
> Tom replied to the previous PR with his thoughts on the matter
> 
> image

It was not a 'meme' when I added it, it was a serious thing for a serious PR.
You still haven't explained how it could be powergamed in a way that can actually happen without admins constantly changing your race.

Helianthus00:
My concerns are more within things like IC race changes, which could be used instead to cycle between beefman and human if the player is banned from that race. It is a fun idea but introduces some fringe issues while not really adding much.

JohnFulpWillard:
> 
> 
> My concerns are more within things like IC race changes, which could be used instead to cycle between beefman and human if the player is banned from that race. It is a fun idea but introduces some fringe issues while not really adding much.

What difference would it be from using mutation toxin at that point? This is harder to do anyways because human mutation toxin is 100% while uh... I guess Lizard mutation toxin? (Felinid/Ethereal mutation toxin isnt a thing outside of the randomized race extract), is only 50% chance.

Helianthus00:
Seems unnecessary to me and would just cause more confusion for players. I'd rather things stay consistent when it comes to bans and such, especially so that the felinid ban can be treated seriously and not as a joke for players.

PR #295, opened 2021-07-23
Tech cult

JohnFulpWillard:
## About The Pull Request

Ports tech cult from our upstream, which is an archived repository, and is itself a downstream of an archived repository.
:)


Helianthus00:
I really wish you had done the beef cytology stuff in another pr, because tech cult is a serious balance concern that I'd need to discuss with the other admins before considering remerging. I personally really hate the t4 organs and the idea that this tech is accessible by the station, albeit very rarely, isn't great. I like the spawner as a whole but it does have huge issues.

JohnFulpWillard:
> 
> 
> I really wish you had done the beef cytology stuff in another pr, because tech cult is a serious balance concern that I'd need to discuss with the other admins before considering remerging. I personally really hate the t4 organs and the idea that this tech is accessible by the station, albeit very rarely, isn't great. I like the spawner as a whole but it does have huge issues.

I can do it separately if this doesn't get merged

JohnFulpWillard:
I'll bring this back if admins are fine with it.

PR #301, opened 2021-07-26
Brig Physician's return of Sec office access & Sec lockers require ACCESS_WEAPONS

JohnFulpWillard:
## About The Pull Request

We can give Brig Physicians their previous access back now!

There was a paragraph of text saying the flaws of this PR, but it is fixed and there isn't any problems anymore. We're good to go.


fulp_modules/Z_edits/container_init/closet_init.dm
```diff
@@ -0,0 +1,44 @@
...
+	new /obj/item/clothing/shoes/jackboots/digitigrade(src)
+	. = ..()
+
+/obj/structure/closet/secure_closet/captains/Initialize()
+	new /obj/item/clothing/shoes/brown/digitigrade(src)
+	. = ..()
+
+/// Security locker ID edit - Changes it to ACCESS_WEAPONS so Deputies no longer have access to it
+/obj/structure/closet/secure_closet/security
+	req_access = list(ACCESS_WEAPONS)
```
SpaceDragon00, line 28: Since this is an assignment override and not an additional variable,, it takes over the security access and allows bartenders to open sec lockers

SpaceDragon00: If you added the access as part of Initialize it would function correctly, while also extending this change to departmental security lockers

JohnFulpWillard: yeah ok problem?

Helianthus00: I'm fine with bartenders being able to open sec lockers since they don't have access to any of the rooms that contain sec lockers, at least to my knowledge. This is better than deputies having access to lockers they shouldn't.

SpaceDragon00: My suggestion works get trolled so hard 
image
image

PR #303, opened 2021-07-26
Making Selene better

TheOnlySolitaire:
- Moved the crayons in dorms to a more efficient location one table north.

## About The Pull Request

Making it so I can get the crayons from the north table in dorms, not south, for better feng shui.

## Why It's Good For The Game

Antags can get crayons faster and win.

## Changelog
:cl:
add: crayons from southern dorms table to northern.
/:cl:

PR #304, opened 2021-07-26
Revert "Making Selene better"

SgtHunk:
Reverts fulpstation/fulpstation#303

This was a PR merged without taking in account the balance changes. We don't want grudgecode in our codebase.


JohnFulpWillard:
24 hour rule

PR #317, opened 2021-08-03
Literally just disables Perma prisoners

JohnFulpWillard:
## About The Pull Request

'xcept on fulp maps that have prison restaurant :)

## Why It's Good For The Game

Prisoners are worse greytiders than assistants as they have literally nothing to do, how are 4 prisoners meant to share one console of Orion Trail? And they cant escape either, not without reason to, anyway.


JohnFulpWillard:
Job slots are part of config so this is a config edit rather than a code one.

PR #319, opened 2021-08-05
Curses Bloodsucker code

JohnFulpWillard:
## About The Pull Request

WHY IS THIS WORKING
WHY IS ANY OF THIS WORKING

I WANTED TO SEE HOW FAR I CAN BREAK BLOODSUCKER CODE
AND IT WON'T BREAK

hhhhhhhhhhhh ----------
- Makes powers use signals
- Makes Frenzy a component
- Turns a ton of Bloodsucker messages into balloon alerts
- Merges SpendRank and SpendVassalRank
- Fixes some stuff that I really should've done in a separate PR - Includes Right Clicking shit
- Adds coffin unclaiming via AltClick
- Changes how Bloodsucker object descriptions work
- Removes tool requirement for Metal coffins
- Cloak of Darkness now uses the same animation as ninja suit rather than its own thing
- Rebalances Power's blood costs (mostly to make up for moving to signals instead of specific timers)
- Adds an examine to Coffins to the person that claimed them, to help them with previously hidden mechanics.


fulp_modules/main_features/bloodsuckers/code/powers/trespass.dm
```diff
@@ -12,9 +12,8 @@
 	var/turf/target_turf // We need to decide where we're going based on where we clicked. It's not actually the tile we clicked.
 
 /datum/action/bloodsucker/targeted/trespass/CheckCanUse(display_error)
-	. = ..()
-	if(!.)
-		return
+	if(!..())
```
SpaceDragon00, line 15: This change feels unnecessary, it makes the code less readable for a very negligible benefit. I think it'd be better if this part was reverted 
```suggestion
	. = ..()
	if(!.)
```

JohnFulpWillard: All other powers do this so I can changed it for consistency

fulp_modules/main_features/bloodsuckers/code/bloodsucker/bloodsucker_life.dm
```diff
@@ -263,24 +260,25 @@
...
-	/// Nutrition - The amount of blood is how full we are.
+	// Nutrition - The amount of blood is how full we are.
 	owner.current.set_nutrition(min(owner.current.blood_volume, NUTRITION_LEVEL_FED))
 
 	// BLOOD_VOLUME_GOOD: [336] - Pale
 //	handled in bloodsucker_integration.dm
+
+	// BLOOD_VOLUME_EXIT: [250] - Exit Frenzy (If in one) This is high because we want enough to kill the poor soul they feed off of.
+	if(owner.current.blood_volume >= FRENZY_THRESHOLD_EXIT && Frenzied)
+		qdel(owner.current.GetComponent(/datum/component/bloodsucker_frenzy))
```
SpaceDragon00, line 271: Wouldn't it make more sense to simply handle all frenzy code inside of the frenzy component, rather than adding and qdel-ing the component whenever it's needed?

fulp_modules/main_features/bloodsuckers/code/powers/_powers.dm
```diff
@@ -185,36 +191,42 @@
...
 			return
-		bloodsuckerdatum.AddBloodVolume(-bloodcost)
-	else
 		var/mob/living/carbon/human/H = owner
 		H.blood_volume -= bloodcost
 
 /datum/action/bloodsucker/proc/ActivatePower()
 	if(amToggle)
-		UsePower(owner)
+		RegisterSignal(owner, COMSIG_LIVING_BIOLOGICAL_LIFE, .proc/UsePower)
```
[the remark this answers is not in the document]
JohnFulpWillard: It isn’t, I rebalanced the constant cost of powers to make up for it, but in general powers cost more now.

fulp_modules/main_features/bloodsuckers/code/powers/trespass.dm
SpaceDragon00: Id prefer if it was consistently . = ..(), but it's not a huge deal either way

fulp_modules/main_features/bloodsuckers/code/powers/_powers.dm
```diff
@@ -49,17 +50,22 @@
...
+//	var/not_bloodsucker = FALSE
+
 /// Modify description to add cost.
 /datum/action/bloodsucker/New()
 	if(bloodcost > 0)
 		desc += "<br><br><b>COST:</b> [bloodcost] Blood"
-	if(warn_constant_cost)
-		desc += "<br><br><i>Your over-time blood consumption increases while [name] is active.</i>"
+	if(constant_bloodcost)
+		desc += "<br><br><b>CONSTANT COST</b><i>[name] costs [constant_bloodcost] Blood maintain active.</i>"
```
JohnFulpWillard, line 68: shit

PR #321, opened 2021-08-06
Bloodsucker crypt code ?improvements?

JohnFulpWillard:
## About The Pull Request

what the fuck am I doing

Was part of #319 but I've changed so much that I don't want it to be in the same PR - So this PR requires that one to be merged first, and this one still needs more work & testing.


JohnFulpWillard:
I’ll bring it back when I’m back

PR #322, opened 2021-08-07
(Most of) Command/Security can become Midround Bloodsuckers

JohnFulpWillard:
## About The Pull Request

They become a Bloodsucker through a Vampiric Accident, like how anyone can roll obsessed, anyone should be able to roll Bloodsuckers, minus some people that would be too OP to have.

### DIFF
--- a/fulp_modules/main_features/bloodsuckers/code/rulesets/bloodsucker_rulesets.dm
+++ b/fulp_modules/main_features/bloodsuckers/code/rulesets/bloodsucker_rulesets.dm
@@ -56,7 +56,12 @@
 	name = "Vampiric Accident"
 	antag_datum = /datum/antagonist/bloodsucker
 	antag_flag = ROLE_BLOODSUCKER
-	protected_roles = list("Captain", "Head of Personnel", "Head of Security", "Warden")
+	protected_roles = list(
+		// Command
+		"Captain", "Head of Personnel", "Head of Security",
+		// Security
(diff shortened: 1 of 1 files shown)


SpaceDragon00:
Just from a balance standpoint I think Sec Officer bloodsuckers would be way too easy, especially since many people are afraid to interfere with an "arrest" out of fear of being bwoinked. This isn't a problem with most other conversions, including vassalization, since all of sec can see they arent mindshielded, but I think the tools they get, the access they have, and the power they inherently hold via the rules would make this too easy to be fun for anyone, even the person playing

Im more fine with Dets and Brig Physes since theyre not supposed to be arresting anyways

Helianthus00:
I wish there was some way to make these roles able to get it but at a much lower weight. I don't want command/sec bloodsuckers to be common but I think it is a good idea to make it possible. Could you make a second event for only them that is less common?

JohnFulpWillard:
> I wish there was some way to make these roles able to get it but at a much lower weight. I don't want command/sec bloodsuckers to be common but I think it is a good idea to make it possible. Could you make a second event for only them that is less common?

The event only rolls max once per round, making it a second event would mean both the events could happen once per round, meaning 2 midround vamps at most

Helianthus00:
After discussing this with the admin team, we're ok with the command being bloodsuckers but don't want to see security kept in here. If you cut it down to just command, this would work.

PR #329, opened 2021-08-12
[TM ONLY] Hotfix to circuit component options

JosephJomama:
## About The Pull Request

Fixes a bug where you can't select from circuit dropdown menus. Identical to https://github.com/tgstation/tgstation/pull/60718

## Why It's Good For The Game

bug bad

## Changelog
:cl:
fix: Fixed circuit component options
/:cl:

PR #334, opened 2021-08-15
[MDB IGNORE] [IDB IGNORE] August TGU

JohnFulpWillard:
## About The Pull Request

Modular stuff :)
Deputy stuff :)
and uh
ye

Closes #329 
Fixes IceBox service
(I think) makes Sec bans now ban all Deputies
We don't actually have MapDiffBot enabled as far as I'm aware, but putting it in just in case.


AffectedArc07:
`[IDB IGNORE]` would also be appreciated. It took 3 minutes, all while the CPU was pegged to the max.

Helianthus00:
> `[IDB IGNORE]` would also be appreciated. It took 3 minutes, all while the CPU was pegged to the max.

Sure, I wasn't aware this was a thing. I'll be sure to let the contributors know.

PR #359, opened 2021-09-20
Moves bodycamera to armor vests

JohnFulpWillard:
## About The Pull Request

Bodycameras get installed in the armor vest instead of the jumpsuit

## Why It's Good For The Game

Antags dont have to disable sec's suit sensors to disable their body camera.


Nobelium-XVIII:
Gives it to all armor vest subtypes like a boss

JohnFulpWillard:
> Gives it to all armor vest subtypes like a boss

Isn’t that the point

PR #389, opened 2021-10-19
Bloodsuckers - Guardians Update

JohnFulpWillard:
## About The Pull Request

Re-adds Bloodsucker Guardians, this time modular and actually protects the Bloodsucker finally.

### DIFF
--- a/fulp_modules/main_features/bloodsuckers/code/bloodsucker/bloodsucker_guardian.dm
+++ b/fulp_modules/main_features/bloodsuckers/code/bloodsucker/bloodsucker_guardian.dm
@@ -1,54 +1,67 @@
-/////////////////////////////////////////////////////////////////
-//          Currently not in game due to being broken          //
-/////////////////////////////////////////////////////////////////
+///Spawn the Guardian if a Bloodsucker uses te guardian creator
+/obj/item/guardiancreator/spawn_guardian(mob/living/user, mob/dead/candidate)
+	var/list/guardians = user.hasparasites()
+	if(guardians.len && !allowmultiple)
+		to_chat(user, span_holoparasite("You already have a [mob_name]!"))
+		used = FALSE
+		return
+	if(IS_BLOODSUCKER(user))
+		var/pickedtype = /mob/living/simple_animal/hostile/guardian/punch/timestop
+		var/mob/living/simple_animal/hostile/guardian/punch/timestop/bloodsucker_guardian = new pickedtype(user, theme)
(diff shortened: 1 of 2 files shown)


SgtHunk:
Are we sure this is something we want to keep/add? The funny jojo reference wasn't even part of the original bloodsuckers design, but rather added by ErwinSnowden (Mynameajeff/VEGA#7059) on https://github.com/TheSwain/Fulpstation/pull/106 (just to further emphasize this, the character he usually played was 'Dio Brando' (This is the same person who added the Dio Brando costume to the autodrobe))

It does seem fun as a nice easter egg, even if a bit out of the scope of bloodsucker source material. Although the PR combined with the person who added it does make it seem like it was added just so that one person could LARP as DIO.
I'd personally be more fine with it if it wasn't a 100% chance (you don't get The World just because you're a vampire. Remember Cream?) or if the chance scaled with vampire level.

This also removes the holoparasite being weaker to damage because timestop is OP as an ability.

JohnFulpWillard:
> 
> 
> Are we sure this is something we want to keep/add? The funny jojo reference wasn't even part of the original bloodsuckers design, but rather added by ErwinSnowden (Mynameajeff/VEGA#7059) on [TheSwain/Fulpstation#106](https://github.com/TheSwain/Fulpstation/pull/106) (just to further emphasize this, the character he usually played was 'Dio Brando' (This is the same person who added the Dio Brando costume to the autodrobe) (Just look at the Ban DB lmao https://centcom.melonmesa.com/viewer/view/mynameajeff))
> 
> It does seem fun as a nice easter egg, even if a bit out of the scope of bloodsucker source material. Although the PR combined with the person who added it does make it seem like it was added just so that one person could LARP as DIO. I'd personally be more fine with it if it wasn't a 100% chance (you don't get The World just because you're a vampire. Remember Cream?) or if the chance scaled with vampire level.
> 
> This also removes the holoparasite being weaker to damage because timestop is OP as an ability.

I do not know who Cream is, but I made it an 80% chance now. Getting a Power miner in itself is rare, so I don't want it to be TOO low of a chance.

JohnFulpWillard:
> 
> 
> > I do not know who Cream is
> 
> Cream is the stand of Vanilla Ice, one of the Vampires in Part 3
> 
> This seems better now, will it trigger on ALL guardian creators or will it only trigger on tarot cards?

All, tarot is only through wizard, so it'll be dumb to limit to just that.

PR #403, opened 2021-11-16
Re-cables the supermatter.

JohnFulpWillard:
## About The Pull Request

Replaces all cables for Selene's supermatter, removing the admin-only power teleporters which only makes repairing the SM nearly impossible and confuses the fuck out of literally everyone.
This was originally meant to be bigger but I got lazy, so I also had removed some spots where 4 walls were cornering eachother, and replaced the Brig maint door with the same access the windoor leading up to it used, as well changed the delivery windoor to security office instead of armory, because it's not the damn armory.

## Why It's Good For The Game

Selene sucks.


JohnFulpWillard:
Fuckers caused a merge conflict on my PR to block it from getting merged

PR #415, opened 2021-11-28
Misdirection can no longer break the laws of physics

FernandoJ8:
## About The Pull Request

This PR fixes the fact Joel's Gun's Misdirection ability could ~~put people inside objects, including but not limited to: 
scrubbers, vents, rolling chair and the nuke, from which they could not exit without admin intervention.
As far as I can tell, the actual intended part remains functionally the same as it used to be.~~
merge people together now, apparently?
This should fix that. 

## Why It's Good For The Game

Fixes a ~~pretty big~~ problem with the thing, hopefully without creating any larger ones. 

## Changelog

:cl:
fix: Joel's Gun's misdirection now uses ~~.throw_at~~ step_towards rather than .Move to avoid previous glitches.
/:cl:

### DIFF
--- a/fulp_modules/features/lisa/joelgun.dm
+++ b/fulp_modules/features/lisa/joelgun.dm
@@ -191,7 +191,8 @@
 				span_hear("You hear aggressive shuffling!"),
 				COMBAT_MESSAGE_RANGE,
 			)
-			main_victims.Move(target)
+			main_victims.throw_at(target, 1, 1, user, FALSE)
+			main_victims.Knockdown(1 SECONDS)
 	// Everyone else will just notice it
 	for(var/mob/living/extra_victims in viewers(5, target))
 		extra_victims.face_atom(target)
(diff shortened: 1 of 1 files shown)


FernandoJ8:
braindead braindead, I'm brainded la-la-la

FernandoJ8:
After a recent bug report in #ask-mods it looks like this might be needed after all

fulp_modules/features/lisa/joelgun.dm
```diff
@@ -191,7 +191,8 @@
 				span_hear("You hear aggressive shuffling!"),
 				COMBAT_MESSAGE_RANGE,
 			)
-			main_victims.Move(target)
+			main_victims.throw_at(target, 1, 1, user, FALSE)
```
JohnFulpWillard, line 194:
```suggestion
			step_towards(main_victims, target)
```

FernandoJ8: Will do, boss.

PR #426, opened 2021-12-04
Fixed outdated radiation flavortext

maxspells:
Changed outdated description of potassium iodide left over from the old rad system so it describes what it actually does

:cl:
del: Outdated potassium iodide description
add: accurate description
/:cl:


FernandoJ8:
Good catch!

PR #429, opened 2021-12-05
Changeling Blood Spider nerf

JohnFulpWillard:
## About The Pull Request

1) You require a reference of a Spider (spider eggs, spiderling) to attempt to make Blood spiders. This makes bagchecking during blood spiders a viable way to figure out who is making them. (The Chef produce console sells spider eggs)
2) You must stand still to make Spiders, rather than make them instantly (15 seconds).
3) There's a random chance to cause a CentCom announcement to your location, with the chance increasing the more spiders you make.

## Why It's Good For The Game

This ability is terribly OP, only held back by Changeling's low weights. In my opinion, this single ability is the largest contributor as to why changelings are so rare in the first place, so nerfing it could allow is to make changelings more common, since it is a player and ironmonger favorite antag to play and fight against.


fulp_modules/Z_edits/antag_edits/changeling_spiders.dm
```diff
@@ -0,0 +1,50 @@
...
+		return FALSE
+
+	user.balloon_alert(user, "creating blood spider...")
+	if(!do_after(user, CHANGELING_BLOODSPIDER_CREATION_DELAY, user, NONE, TRUE, extra_checks = CALLBACK(src, .proc/has_spiders, user)))
+		user.balloon_alert(user, "spider creation interrupted!")
+		return FALSE
+	user.balloon_alert(user, "spider created!")
+
+	if(prob(CHANGELING_BLOODSPIDER_ANNOUNCEMENT_CHANCE*ling_datum.spiders_created) && ling_datum.spiders_created >= 1)
+		priority_announce("We detect a set of strange spiders appearing out of [get_area(get_turf(user))], please dispatch a security force to investigate.", "Central Command")
```
Helianthus00, line 44: The other nerfs are pretty substantial already, I don't think we need this as well.

Helianthus00 (changes requested):

fulp_modules/Z_edits/antag_edits/changeling_spiders.dm
JohnFulpWillard: I find it fair that making a free self healing murderbone antag will reveal your current location, not like the changeling can’t just move afterwards.

Helianthus00: Could also be confused with the other spiders though. I still think it should be removed.

Helianthus00 (approved):

PR #433, opened 2021-12-08
[IDB IGNORE] [MDB IGNORE] December TGU

JohnFulpWillard:
## About The Pull Request

It's been a while since we've had one, we had to deal with a ton of bugs for a while, I think it's time it comes to an end.
TG is currently on a feature freeze so we don't (shouldn't) have to worry about new bugs coming up from this.

This is late because I've spent a ton of time improving Bot, beefmen and bloodsucker code.


JohnFulpWillard:
IDB is failing and I hope it wasn't because of modsuits, I'm IDB ignoring just in case.

PR #434, opened 2021-12-12
Create emergency_xeno shuttle

TheOnlySolitaire:
## About The Pull Request

New shuttle based on an pre-existing shuttle that has been overrun by Xenos. 
The end-goal is to have the shuttle be unlocked by scanning a xenos organ and/or by purchasing the corresponding xeno research. 

## Why It's Good For The Game

New shuttle adds to options and fun in the game.

## Changelog

:cl:
add: Adds new shuttle
/:cl:


SgtHunk:
How does this interact with shuttle-stalling antags, now that because of the shuttle we have 10+ facehuggers that will make player-controlled xenos

TheOnlySolitaire:
> How does this interact with shuttle-stalling antags, now that because of the shuttle we have 10+ facehuggers that will make player-controlled xenos

Very good point. I think Shuttle Hijack could be considerably harder, though a clever antag might use this shuttle to assist in preventing other crew getting onboard so they just need to focus on the hijacking and surviving part. But for sure preventing crew from escaping would become easier. 

Something to consider though is that there are the cult and doom arena shuttles, thats off the top of my head and there may be a couple more, which also have this issue. Is this that far different? 
Would balancing the number of possible player controlled mobs/facehuggers make it more agreeable?

StoneMason:
If there's a blob, revolution or even a smart cult, the shuttle will arrive and self-antaggers/dumbasses will sprint onto the shuttle and get facehugged. Whilst it probably wouldn't be a huge issue as captain's rarely go for cool shuttles, this shit will definitely cause a xeno outbreak a few times. Just make it like the lavaland ruin so it's xenomobs and popped eggs.

SgtHunk (changes requested):
Remove the xeno eggs entirely. A shuttle shouldn't be bringing an antag onto the station. It can include enemies a la cult shuttle, but having people get facehugged and bring xenos is just too much.

wraith-54321:
1 egg in a command only area could work, I dont know, or you could just give the xeno mobs a fun balloon like in the cult shuttle

TheOnlySolitaire:
I will remove the eggs and just leave mobs, is it ok for those mobs and at least for now, remove the sentience balloon to let them be player controlled. If thats ok and requested I'll re-add it but eggs will be gone on next commit.

Helianthus00:
This seems a little odd to just have available to order at any time. If you lock it behind something like emag or xeno research, I'd be more willing to consider this.

fulp_modules/mapping/shuttles/map_templates.dm
```diff
@@ -45,3 +45,14 @@
 	name = "Selene Station Emergency Shuttle"
 	credit_cost = CARGO_CRATE_VALUE * 4
 	description = "The standard-issue escape shuttle for models of station this large. Will get you home in moderate style."
+	
+/datum/map_template/shuttle/emergency/fulp/xeno
+	prefix = "fulp_modules/mapping/shuttles/evac/"
```
JohnFulpWillard, line 50: delete this line

TheOnlySolitaire:
> This seems a little odd to just have available to order at any time. If you lock it behind something like emag or xeno research, I'd be more willing to consider this.

It was always planned to be, I just needed to figure it out. I've updated the file so that it now has:

- locked the shuttle behind alien tech
- removed Queen and Praetorian
- removed all xeno eggs
- removed sentience balloon
- tweaked xeno mobs so there are 7 total, 3 drones, 3 sentinels and 1 non-hostile larva

PR #441, opened 2021-12-19
Beefmen brain trauma is now a preference

JohnFulpWillard:
## About The Pull Request

Beefmen trauma is now a preference that you can choose
Also removes the deadchat one, since it was too good to keep now that it isn't just random.

## Why It's Good For The Game

some beefmen dont want hypnotism every round. some might, idk.

### DIFF
--- a/fulp_modules/features/species/beefmen/beefman_preferences.dm
+++ b/fulp_modules/features/species/beefmen/beefman_preferences.dm
@@ -90,3 +90,21 @@
 
 /datum/preference/choiced/beefman_mouth/apply_to_human(mob/living/carbon/human/target, value)
 	target.dna.features["beefmouth"] = value
+
+//Trauma
+/datum/preference/choiced/beefman_trauma
+	savefile_key = "feature_beef_trauma"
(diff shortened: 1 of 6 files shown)


Helianthus00:
I'm not sure if I like the idea of beefman traumas being selectable. When I discussed this with Wind, he agreed. Though we both think it would be cool if the trauma was randomized rather than being a preference.

JohnFulpWillard:
> I'm not sure if I like the idea of beefman traumas being selectable. When I discussed this with Wind, he agreed. Though we both think it would be cool if the trauma was randomized rather than being a preference.

I preferred it randomized, but I really don't find it fair.
Every other race has all their traits as options, even Lizards can choose to be digitigrade or not.
There's also cool traumas like hypnotic stupor, that many people (basically all beefmen mains) tell me they hate and wants removed.
There's no good traumas in the list and it's impossible to not have one, I thought this was the best way to have as much fun with the species without being forced into shit you never wanted.g

Helianthus00:
I still don't love this solution but we can try it out and maybe rework it down the line.

PR #442, opened 2021-12-24
L

JohnFulpWillard:


StoneMason:
where i charge withing

Nobelium-XVIII:
withig

SgtHunk:
That is not an answer, and I absolutely despise when people use this argument as one.
"ooh, don't worry it's admin only" - well sorry bud, admins abuse things too. They're just ordinary people, put into positions of power - assuming that they're somehow more responsible than everyone else is foolish at best, and ignorant at worst. And it's not a baseless claim either - I've seen them abuse tools given to them more than once, be it by putting themselves in a body of a bloodcrawling, shadowcrawling, fully-armored changeling, or by spawning in a literal clown-ops team.

But disregarding my "mild distaste" for administrative staff, because I'm going mightily off-topic here.
This isn't a bad PR, m'kay? It is good, but it simply doesn't fit here. It'd be a good thing to add to TG (please don't no one will revert it here), Skyrat, etc. But it's not a good thing to have HERE, on a ROLEPLAY server. Yes, it's "admin bus" exclusive, but it's still not a good thing roleplay-wise. And if it isn't, and admins shouldn't spawn it in - then what's even the point?

wraith-54321:
The game was literally unplayable before this fix, 10/10 would grammar again

HugoOdaX:
?

PR #445, opened 2021-12-25
Fixes Ventrue bloodsucker's levelling up with Blood

JohnFulpWillard:
## About The Pull Request

I keep forgetting to fix this, so I'm doing this now that it's still on my mind.


JohnFulpWillard:
added this to https://github.com/fulpstation/fulpstation/pull/446

PR #464, opened 2022-01-10
Brings back Summon Dance Floor

Horatio22:
## About The Pull Request

Exactly what it says on the tin. Brings back Summon Dance Floor from the deleted devil game mode. It may need to have a longer cooldown because of the smoke, but that hopefully won't be an issue. Here's a pic for people who weren't around before it was removed: 
image

## Why It's Good For The Game

Since we changed our policy for wizards, I thought it would be fun to give them more gimmick-y type spells that don't necessarily do anything other than maybe be annoying. 

## Changelog

:cl:
add: Adds Summon Dance Floor to the wizard spellbook

/:cl:


Helianthus00:
fulp_spellbook.dm isn't in the dme

Horatio22:
> fulp_spellbook.dm isn't in the dme

image

PR #477, opened 2022-01-19
Fixes the helio bar stools facing the wrong way

HugoOdaX:
## About The Pull Request
The stools are facing the opposite direction, unless you change them manually, the bar bots won't take drinks from the tables. And that's bad for new bartenders, as they may not notice that the stools are facing the other way.

## Why It's Good For The Game
SERVICE WINS AGAIN


HugoOdaX:
> The most nit-picky PR to ever exist.

how is it nitpicky? literally prevents newer bartender from using the bots unless they figure out how to rotate the stools

StoneMason:
People actually serve bots? I thought that was a myth.

PR #481, opened 2022-01-23
Deputies: Lockers & Beret destroy objectives

JohnFulpWillard:
## About The Pull Request

Adds Deputy lockers, holding the beret and some other items

Maybe in the future we can add more to it so deputies can have things to do, like a buildable machine of some sort.

Adds objectives to destroy their berets, since they are rather strong and is the Deputy unique item.


fulp_modules/Z_edits/antag_objectives.dm
```diff
@@ -0,0 +1,32 @@
...
+
+/datum/objective_item/steal/low_risk/deputy_beret_sup
+	name = "the supply deputy's beret"
+	targetitem = /obj/item/clothing/head/fulpberet/supply
+	excludefromjob = list(JOB_DEPUTY_SUP)
+	exists_on_map = TRUE
+
+/datum/traitor_objective/destroy_item/low_risk/New(datum/uplink_handler/handler)
+	. = ..()
+	possible_items |= list(
```
SpaceDragon00, line 27: Do not do it like this. The normal low_risk list is evenly weighted with 3 possible items; filling it with 4 different berets means the steal objective will be a beret 4/7 times. At most, have a generic beret objective that generates a random beret to steal when it is rolled

JohnFulpWillard: ok????

JohnFulpWillard: We want them to destroy deputy berets so I don't see a problem with this.

SpaceDragon00: You're oversaturating an otherwise varied objective list with 4 different flavors of "Kill someone with 90% block who has a stun baton." No one would do it, especially not for the TC reward. You can't put this on the same level as stealing a fireaxe

SpaceDragon00: If anything, put this in very_risky, since nobody gives a shit about the blackbox anyways, and at that point the TC reward might incentivize at least some people into attacking a deputy (Probably still wont though)

JohnFulpWillard: stealing a fireaxe is also hard, as is stealing a shotgun/null rod as the Atmos tech/Bartender/Chaplain likely will not hand it over to you.

Sheits: Deputies are pretty strong however unlike shotgun and null rod objectives, you can just throw a hat at the deputy if theyre wearing it to get it. Its also an opt-in objective.

SpaceDragon00: Out of all of those jobs, only one has a stun baton and 90% block chance. It's not low risk by any means

JohnFulpWillard: bartender's shotgun can stamcrit you faster than a baton can, and has a whole desk (that you get hardstunned if you try to climb over) between you and them. Chaplain has 50% melee resistance on top of their null rod as a weapon. atmos techs have a fucking fireaxe. I don't see the problem.

PR #485, opened 2022-01-24
Organizes all Fulp cargo crates into one folder/file and adds a FNAF plushy crate to cargo

ReinaCoder:
Github has been fucky but this should be good. This is tested and works.
Someone asked for a cargo crate with the fnaf plushies so I'm obliging by their suggestion (finally a good suggestion). 
image
(Sprites by JACK)


JohnFulpWillard:
you didnt delete the brigdoc  file you just deleted it from .dme

PR #487, opened 2022-01-24
Fixes all of the 375 roundstart active turfs on Selene Station

Pepsilawn:
Sometimes when I played on Fulpstation server when they were running their famous map "Selene Station" I would see this:
image
And think to myself: "Wow! This is certainly a 'Selene Station' moment!" but it has now come time for me to fix these issues that make us look bad on the lobby screen

Below are examples of things I fixed there were some more that i didnt screenshot but I think some of these are funny or what do you think

image
image
(solars not using the airless subtypes were like 80% of them)
image
image


SgtHunk:
Who made this shitty map 👎
just remove it tbh

PR #508, opened 2022-02-05
Perma expansion 2

JohnFulpWillard:
## About The Pull Request

- Prisoners all share a Prison budget
- A shuttle comes by every so often with an objective, prisoners have 3-4 minutes to do it.
- These objectives can range from getting a certain xenobio extract, to cooking food, to cleaning a nasty shuttle 
- If they do it, their budget gains money. If they fail, they lose money.
- They can spend money with the budget directly at cargo

Just as bonuses:
- There's speakers that broadcast this over perma
- Prison chef now has a leaderboard to keep track of score for their restaurant bots
- General maintaining of current restaurant bot code (like the outdated food list)
- Splits some files into separate ones so it doesn't look terrible.


dnworden:
Playtested this today, it looks great! However, I have a few questions/suggestions:

Where is the prison budget? The console doesn't seem to let prisoners actually order anything, only request stuff. It's also not adding anything to prisoner's personal accounts when the shuttle objectives are completed.

Prison shuttles are really interesting, but some stuff should be organized a bit better. Prisoners can just print knives out of the dinnerware vendor on the chef shuttle, and get excessive amounts of alcohol bottles out of the bar shuttle. Now, prisoners are going to get knives and booze anyway sooner or later, but maybe the chef shuttle should have just one knife (perma already has contraband for other knives) and one rolling pin, and the bar shuttle should have alcohol dispensers and one of each bottle of alcohol plus the dispensers instead of having a full Booze-O-Mat.

Even though there doesn't seem to be a way to spend the prison budget yet, I also noticed that failing to complete a request penalizes the prisoners by reducing the budget. Maybe the penalty should be removed? Prisoners might have other things that they need to do like farming or cooking or making license plates, and they shouldn't be penalized just because they were busy working elsewhere.

Those three points aside, I'm really excited for this PR. As an avid prisoner player myself, thank you for making this and having so much content!

JohnFulpWillard:
> Where is the prison budget? The console doesn't seem to let prisoners actually order anything, only request stuff.

I haven't checked much in it, but it is SUPPOSED to order using the prison budget. If it requires being made a cargo console rather than a request console, I would be dissapointed, because letting prisoners order anything and manage the cargo console is rather annoying. Maybe in the future I can do like a chef produce console-type thing and it arrives on the prison shuttle instead.

> Prisoners can just print knives out of the dinnerware vendor on the chef shuttle, and get excessive amounts of alcohol bottles out of the bar shuttle.

It was what I was hoping for, letting prisoners get otherwise contraband mid-round, rather than all roundstart, means security will have to constantly make sure prisoners don't have contraband.

> I also noticed that failing to complete a request penalizes the prisoners by reducing the budget. Maybe the penalty should be removed? Prisoners might have other things that they need to do like farming or cooking or making license plates, and they shouldn't be penalized just because they were busy working elsewhere.

I'd be fine with that, all budgets get 500 credits every payday so any money lost from the shuttles is mostly put into negating that, and losing money for not doing what is asked of you should still have consequences.

PR #522, opened 2022-02-12
makes job positions a code thing rather than a config one.

JohnFulpWillard:
## About The Pull Request

From this conversation in the discord:
image

Job positions here have been copy pasted from live fulp servers' config. I will not be editing any job positions in this PR, that should be done at a later date after discussing what job slots to edit. Prisoners will likely be the first one edited after this.

## Why It's Good For The Game

random epic trolling where we have no jobs open
also fulp maps, that actually has stuff for prisoners to do, will be able to have roundstart permas, and we can remove them from other maps.


fulp_modules/Z_edits/job_edits.dm
```diff
@@ -0,0 +1,131 @@
...
+ *	# Job edits
+ *
+ * 	Overwrite TG code to add exp requirement of a few dangerous jobs.
+ * Additionally setting fulp config job slots, since TG's job config is unstable, and overwrites map .json job positions.
+ */
+
+/** Security */
+/datum/job/warden
+	total_positions = 1
+	spawn_positions = 3
```
Helianthus00, line 11:
```suggestion
	spawn_positions = 2
```
This is how the current config is

Helianthus00 (changes requested):

Helianthus00 (approved):

PR #530, opened 2022-02-18
solitiare arrivals re-make

JohnFulpWillard:
## About The Pull Request

- Addresses some problems brought up in https://github.com/fulpstation/fulpstation/pull/488
- Remakes most of arrivals, including holodeck/dorms/maints around it
- Moves science down
- Fixes random problems I found while working on this
- Cuts back on more maints

image

## Why It's Good For The Game

Solitaire needs a complete disposal overhaul, but first we need to overhaul all the individual departments.
Service, Medical, Engineering and Supply, and Departures is what's left to be done after this PR, as well as maints but I'd worry about that last.
We can make Solitaire an actual unique map since it does have a lot of promise to it. I've also sunk to much time into this map to just let it die.

### DIFF
--- a/_maps/map_files/SolitaireStation/Solitairestation.dmm
+++ b/_maps/map_files/SolitaireStation/Solitairestation.dmm
@@ -65906,7 +65906,6 @@
 /area/maintenance/starboard/fore)
 "urx" = (
 /obj/structure/lattice,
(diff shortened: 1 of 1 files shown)


Pepsilawn:
does sci maints have 2 apcs

PR #533, opened 2022-02-19
beef statio nimprovement

Pepsilawn:
replaces some of the titanium walls with the nodiagonal variants because they're somewhat ugly in some places see below
image
some other fixes like the space injector turned into a passive vent, air alarms & the bar tables having the wrong dirs & some qol stuff like being able to upload books into the archive

## Changelog

:cl:
image
/:cl:


JohnFulpWillard:
is that jed

Pepsilawn:
yea

Greniza:
in before putting an image in the CL bricks the changelog

Pepsilawn:
what changelog

Ebb-Real:
such a statio ninprovement man

PR #534, opened 2022-02-19
removes lattice

Pepsilawn:
image
image
image
Yes these purple parts are indeed walls that are filtered out in strongdmm we have so many fucking lattices in our walls its incredible its ruining my gameplay why do we have so many lattices inside our walls even tg maps have them they're not visible or anything they're not intended yet no one's doing anything about it whats up with that

### DIFF
--- a/_maps/map_files/Heliostation/Heliostation.dmm
+++ b/_maps/map_files/Heliostation/Heliostation.dmm
@@ -67016,6 +67016,11 @@
 /obj/machinery/exodrone_launcher,
 /turf/open/floor/plating,
 /area/cargo/warehouse)
+"hHB" = (
+/obj/structure/lattice,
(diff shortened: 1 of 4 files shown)


SgtHunk:
just remove the maps tbh

Pepsilawn:
> just remove the maps tbh

its on tg too i saw like 6 of them on meta alone

JohnFulpWillard:
man i wish this was merged

JohnFulpWillard:
if it gets merged NOW though then it will conflict with TGU and I will be like >:(

JohnFulpWillard:
this doesnt even remove them all ⛽

Pepsilawn:
i forgot i pr'd this here after i went to tg to fix theirs

PR #540, opened 2022-02-23
Makes cloak not care for people who can't see you

JohnFulpWillard:
## About The Pull Request

Cloak of darkness now only takes into account people who can visibly see you, rather than people near you.

## Why It's Good For The Game

People can't see through walls, unless they can, in which case they can, but generally can't.


Greniza:
> People can't see through walls, unless they can, in which case they can, but generally can't.

🤨

PR #546, opened 2022-03-05
Adds linter check to deter from bad mapping practices

Pepsilawn:
all in the title


JohnFulpWillard:
so true....

OperativeLyn:
so true....

Pepsilawn:
some asshole in a labcoat just came to my house and punctured all of my car's tires, closing this while im helping the police with the investigation

PR #555, opened 2022-03-13
FIXES meatman limb duplication glitch

SmoSmoSmoSmok:
## About The Pull Request

 meatmen can no longer infinitely generate meat when sawing off their limbs

## Why It's Good For The Game

 i found another unrelated  bug where beefmen can stand on 0 legs after they saw off their legs. 
HAPPENS at random IDK whats causing it
## Changelog

:cl:

/:cl:

### DIFF
--- a/fulp_modules/features/species/beefmen/mob/beefman.dm
+++ b/fulp_modules/features/species/beefmen/mob/beefman.dm
@@ -257,9 +257,9 @@
 		span_notice("Your [affecting.name] pops right off."))
 	playsound(get_turf(user), 'fulp_modules/features/species/sounds/beef_hit.ogg', 40, 1)
 	// Destroy Limb, Drop Meat, Pick Up
-	var/obj/item/dropped_meat = affecting.drop_limb()
+	var/obj/item/food/meat/slab/dropped_meat = affecting.drop_limb()
 	//This will return a meat vis drop_meat(), even if only Beefman limbs return anything. If this was another species' limb, it just comes off.
-	if(istype(dropped_meat, /obj/item/food/meat/slab))
+	if(dropped_meat)
 		user.put_in_hands(dropped_meat)
 	return TRUE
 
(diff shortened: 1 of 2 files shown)


OneAsianTortoise:
Smokey nerfing himself

SmoSmoSmoSmok:
> Smokey nerfing himself

I need TO give the current generation some fighting chance against me

SpaceDragon00:
> i found another unrelated bug where beefmen can stand on 0 legs after they saw off their legs.

> HAPPENS at random IDK whats causing it

I fixed this bug when it was happening to digi lizards, but knowing Beefmen there's probably some snowflake code that is causing it to resurface. Or who knows maybe my fix wasn't as universal as I intended it to be

The original problem came from the "number of legs" variable (Whatever its called) being less than 0, or somehow more than 0 when they had 0 legs. If I had to guess, some part of Beefman code manually adjusts that variable instead of using the method that ensures it doesn't go below 0

OneAsianTortoise:
> > Smokey nerfing himself
> 
> I need TO give the current generation some fighting chance against me 

We should up to 2% to 50% then

SmoSmoSmoSmok:
> > i found another unrelated bug where beefmen can stand on 0 legs after they saw off their legs.
> 
> > HAPPENS at random IDK whats causing it
> 
> I fixed this bug when it was happening to digi lizards, but knowing Beefmen there's probably some snowflake code that is causing it to resurface. Or who knows maybe my fix wasn't as universal as I intended it to be
> 
> The original problem came from the "number of legs" variable (Whatever its called) being less than 0, or somehow more than 0 when they had 0 legs. If I had to guess, some part of Beefman code manually adjusts that variable instead of using the method that ensures it doesn't go below 0

OK i fixed it. num_legs was updating correctly but usable_legs wasnt for some reason. the glitch would happen everytime you tore off a disabled limb

PR #559, opened 2022-03-15
Feed rework

JohnFulpWillard:
## About The Pull Request

Makes feed unable to be used at level 0
Removes useless garbage from feed
reworks how its done
You don't spray blood when you stop feeding
You now feel the effect of feeding off of a dead person regardless of whether or not YOU killed them
Reworks how you find a target, too.
Basically, re-does everything. Explanation in-game was also updated to make up for the changes.
Removes the stun, only keeps the part where you put to sleep, which scales with level now

## Why It's Good For The Game

feed is broken, both in terms of balance and in the literal sense that it's fucking broken as shit, so I decided to completely re-do it from the ground up.


fulp_modules/features/antagonists/bloodsuckers/code/powers/feed.dm
```diff
@@ -1,367 +1,230 @@
...
+		bloodsuckerdatum_power.give_masquerade_infraction()
+	else
+		owner.balloon_alert(owner, "feed unnoticed.")
 
-	// FEEEEEEEEED!! //
-	ADD_TRAIT(user, TRAIT_MUTE, BLOODSUCKER_TRAIT) // My mouth is full!
-	ADD_TRAIT(user, TRAIT_IMMOBILIZED, BLOODSUCKER_TRAIT) // Prevents spilling blood accidentally.
+	ADD_TRAIT(owner, TRAIT_MUTE, FEED_TRAIT)
+	ADD_TRAIT(owner, TRAIT_IMMOBILIZED, FEED_TRAIT)
+	. = ..()
```
SuperNovaa41, line 126:
```suggestion
	return ..()
```

PR #574, opened 2022-03-27
[Heliostation] fixes problems with atmos

JohnFulpWillard:
## About The Pull Request

Replaces gas masks with atmos gas masks
Fixes incinerator power to not rely on an incinerator to power it directly
Makes atmos pipes not connected to mix if you try to edit them in any way
Makes atmos pipes easier to add layer manifolds to without plasmaflooding yourself

## Why It's Good For The Game

how isn't it


Pepsilawn:
while ur at it can u fix the camera names incinerator is called something like construction & i think theres like 3 other fucked up ones in the engi hallway idk i played engi deputy once on helio and instantly closed the game after checking the telescreen

PR #592, opened 2022-04-07
Adds a prison shuttle gate

JohnFulpWillard:
## About The Pull Request

Adds a gate to the prison shuttle which prevents anything/anyone not wearing an ID from passing by.
This also prevents you from throwing things like mail out, but it still works to just put it in your bag and leave.
This can also be fixed in the future but I don't see it as too much of a problem to fix currently (If anyone else wants to, my idea is to RegisterSignal something on the item that blocks the gate from letting you leave, which should be put on mail and trash).

## Why It's Good For The Game

No more slime flooding permabrig, and permabrig on maps don't look like shit anymore.


Pepsilawn:
Making the cleaning shuttle job harder even tho it was already impossible 2 beat

SgtHunk (changes requested):
Please double-check the airlock names as it seems you're using the same "Perma Cell 6" airlocks

PR #594, opened 2022-04-12
removes prisoners

JohnFulpWillard:
## About The Pull Request

Prisoners should still spawn on Helio/Selene/Pubby because of config, this removes them from all other maps.


Helianthus00:
For now, we want prisoner slots to remain open.

PR #598, opened 2022-04-12
Adds maps to fulp

Pepsilawn:
## About The Pull Request

Selene arrivals has a map of a version of boxstation from like 10 years ago which is kind of weird I think so I thought it'd be cool to replace it with an actual map of selene & i did the same to helio tbh the sprites are kind of shite but so are the other map sprites

image
image

## Why It's Good For The Game

something about new players


Helianthus00:
Wow this is actually really cool

PR #602, opened 2022-04-14
[MDB IGNORE] April TGU

JohnFulpWillard:
## About The Pull Request

New limbs code (had to make beefmen work for it)
New action buttons code (had to make bloodsuckers work for it)
New access lists (had to make deputies work for it)
A ton of pathing changes that had no updatepaths page to help with, so most was manually done on Helio/Selene/Pubby. Very sad about it still.

## Why It's Good For The Game

hey


FernandoJ8:
Jotting down some of the broken stuff I dug up so you don't have to:
- Plantis can wear digi shoes
- Newscasters and bounty boards got reworked and changed their paths so they're broken on the fulp maps
- Waypointing beacons need to be replaced with navigation landmarks in the fulp maps
- The way the different mech suit colours work was refactored so the paths are broken for all but the red one on selene

PR #603, opened 2022-04-16
Cleanup shuttle drop pods refactor

FernandoJ8:
## About The Pull Request

Refactors the code for the prison cleanup shuttle drop pods to use landmarks, optimizing it a bit in general and fixing the following bugs:

- Drop pods could drop on top of the engine tiles of the pod
- Drop pods could drop in the shipping line hyperspace
- The trash spawners could also spawn items on top of engine tiles because their scatter range was set to 1

## Why It's Good For The Game

Bugs bad, pesticide good

## Changelog

:cl:
fulp changelogs are a social construct
/:cl:


FernandoJ8:
Hm, actually do you reckon that it's better to have that while loop or to just get all the turfs, add them to a list and pick from that list?

JohnFulpWillard:
idk if thsi was causing lag in the past but if it was at least it won't anymore.

FernandoJ8:
> idk if thsi was causing lag in the past but if it was at least it won't anymore.

Probably not a lot, since it's one of the several shuttles that can appear and they come in with about 5 minutes in between each, but any optimization is good. Plus the bugfixes were needed.

PR #607, opened 2022-04-19
Adds beefman cookies

maxspells:
[Beefmen now get slabs of meat instead of a cookie when they have their prayers answered](https://forum.fulp.gg/t/make-beefmen-receive-meat-instead-of-a-cookie-for-answered-prayers/4933)


maxspells:
Oh I don't think John's cookie PR was in this TGU so we'll have to wait until the next one

PR #612, opened 2022-04-21
blood-sucker-tgui - Adds a new tgui interface for the bloodsucker antag

raffclar:
## About The Pull Request

Adds a Bloodsucker TGUI interface by extending tgui with a new "fulpui" package.

## Why It's Good For The Game

Improved UX for the bloodsucker introduction window. Lays the groundwork for future fulpstation interfaces.

Existing Bloodsucker tips interface:
image
New Bloodsucker tgui interface:
image

## Changelog

:cl:
add: Bloodsucker antag tgui interface
/:cl:

### DIFF
--- a/tgui/packages/tgui/routes.js
+++ b/tgui/packages/tgui/routes.js
@@ -81,8 +81,15 @@ export const getRoutedComponent = store => {
   while (!esModule && interfacePathBuilders.length > 0) {
     const interfacePathBuilder = interfacePathBuilders.shift();
     const interfacePath = interfacePathBuilder(name);
-    // Fulp edit - Fulp TGUI
-    esModule = loadInterface(interfacePath);
+    try {
+      // Fulp edit - Fulp TGUI
+      esModule = loadInterface(interfacePath);
+    }
+    catch (err) {
+      if (err.code !== 'MODULE_NOT_FOUND') {
+        throw err;
+      }
+    }
   }
   if (!esModule) {
     return routingError('notFound', name);
--- a/tgui/packages/tgui/routes.js
+++ b/tgui/packages/tgui/routes.js
@@ -8,6 +8,7 @@ import { selectBackend } from './backend';
 import { Icon, Section, Stack } from './components';
 import { selectDebug } from './debug/selectors';
 import { Window } from './layouts';
+import {loadInterface} from "../fulpui-patches";
 
 const requireInterface = require.context('./interfaces');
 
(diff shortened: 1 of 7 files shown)


tgui/yarn.lock
```diff
@@ -4320,6 +4320,24 @@ __metadata:
   languageName: node
   linkType: hard
 
+"fulpui-patches@workspace:packages/fulpui-patches":
```
JohnFulpWillard, line 4323: this being in the tgui folder doesnt bother me, but it still feels like there's a ton of TG edits here. Is this part able to be a separate file or is that not how it works? I'm still very new to TGUI

raffclar: This needs to be kept. The tgui package is dependent on it. I tried to do very some hacky stuff within webpack and it was not worth the mess it made.

tgui/packages/tgui/routes.js
```diff
@@ -57,14 +56,7 @@ export const getRoutedComponent = store => {
   while (!esModule && interfacePathBuilders.length > 0) {
     const interfacePathBuilder = interfacePathBuilders.shift();
     const interfacePath = interfacePathBuilder(name);
-    try {
-      esModule = requireInterface(interfacePath);
-    }
-    catch (err) {
```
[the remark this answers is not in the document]
raffclar: This is the critical bit that loads the component for the player to see and interact with. I'm not sure what you're suggesting exactly but we need two try/catch attempts, one for fulp interfaces and the fallback for the original tg interfaces.

tgui/packages/tgui/routes.js
```diff
@@ -57,14 +56,7 @@ export const getRoutedComponent = store => {
...
-      if (err.code !== 'MODULE_NOT_FOUND') {
-        throw err;
-      }
-    }
+    esModule = loadInterface(interfacePath);
```
JohnFulpWillard, line 59: This is what I mean, only changing the file we need to.
I did this locally and it seems to work:
image

It will make it easier to do future TGU's, especially since we barely know TGUI

```suggestion
    try {
      esModule = loadInterface(interfacePath);
    }
    catch (err) {
      if (err.code !== 'MODULE_NOT_FOUND') {
        throw err;
      }
    }
  }
```

JohnFulpWillard:
Last thing I want to bother you about, we have to distinguish these as TG edits, comments saying ``// Fulp edit - Fulp TGUI``, and adding them to tg_edits.md for TG updates in the future. Wanted to say it last so we can know what will be needed in the end or not.

tgui/packages/tgui/routes.js
raffclar: I see what you mean now. I can do that sure

raffclar: Sorted

raffclar:
I struggled to write tests for this so I'm dropping tests for now. There's no component-level tests upstream and there's something missing config settings. As Fulp tries to minimise merge conflicts it would be a bad idea to try add better test support.

@JohnFulpWillard
* Removed powers tab, removal is kept in a new commit.
* Took your advice and maintained the try/catch blocks.
* Added entries to the tg_edit doc and a few comments where possible.
* Fixed failing CI

PR #618, opened 2022-04-23
[TM First] Bloodsucker vassalization objective refactor and bugfix

FernandoJ8:
## About The Pull Request

Generally refactors the vassalization / formerly protege / currently conversion BS objective and addresses several bugs it had in the process.

Due to the fact it's a conversion antag objective, testing has been tricky and somewhat limited, so this should be test merged to find unspotted runtimes, bugs etc. before merging.

## Why It's Good For The Game

According to testing, this objective was broken in multiple places and the department variation wouldn't work at all. Now they should both work and the code's cleaner in general.

~~You may rightfully wonder why I kept them as subtypes of a parent objective that does literally nothing because, functionally, there is indeed no reason for it. I kept it for the sake of categorization, since they're variations of the same concept and, both before and after this PR, can't coexist (you get one or the other), but I'm not super convinced on that choice.~~
Now it has a reason to be.

## Changelog

:cl:
refactor: refactors the protege (now named conversion) bs objective code
/:cl:


FernandoJ8:
I'm actually braindead, hold on.

FernandoJ8:
oh what in the fuck...
working on multiple computers be dammed, ugh

FernandoJ8:
Why's it even showing those as changes, the commits are already on the repo, it's changing literally nothing

PR #632, opened 2022-05-01
Adds body camera to sec vests

JohnFulpWillard:
## About The Pull Request

Lets sec vests hold body cameras.

## Why It's Good For The Game

security doesn't have to sacrifice better clothes for lack of a body camera (and it has lower armor already so :/ )


SgtHunk:
does this include the justice armor ?

PR #641, opened 2022-05-12
Removes Digitigrade Shoes

Ebb-Real:
## About The Pull Request

Kapulimbs broke digitigrade shoes and anyone can wear digi shoes making it look broken and very cursed. If we wanted to keep them we would need to make special shoes that only fit digitigrade but that would also need to have lizards be able to wear normal shoes as far as i understand which would also look weird. Digi shoes aren't worth the effort and loopholes just to make a few players who can just chose normal legs happy.

## Why It's Good For The Game

No more digitigrade human psychopaths and less things which only affect a very small amount of players to maintain.

## Changelog

:cl:
del: Removes digitigrade shoes.
/:cl:


JohnFulpWillard:
if digi lizards want digi shoes they should code it themselves?

Ebb-Real:
> if digi lizards want digi shoes they should code it themselves?

so true

SpaceDragon00:
This problem is very easily fixable. Why remove it?

1. Define a condition (Variable, bitflag, etc. Something that makes it clear this is only for digitigrade shoes. You could use a custom bitflag or just include a new variable)
2. Evaluate that condition on the objects, either through inheritance (obj/item/shoes/digitigrade/shoenamehere) on a standardized proc or through an element that evaluates the conditions on move or equip

SpaceDragon00:
Refactor it for inheritance if you want to actually improve the code, or make a simple element if you just want to avoid the bug with as little new code as possible. The removal will take you more effort than a fix

JohnFulpWillard:
@SpaceDragon00 wyci

SpaceDragon00:
> @SpaceDragon00 wyci

Bet

Ebb-Real:
> > @SpaceDragon00 wyci
> 
> Bet

yeah sure, i'll close this if you put up a PR

SpaceDragon00:
I changed my mind I'm not gonna be baited into coding a fix out of spite lmao

This just seems silly. A bug was left over from the TGU, so the feature is being removed? That's not exactly a contribution more than it is an excision

SpaceDragon00:
In short, if it's too much effort to fix the bug, why isn't it too much effort to try and remove all the content?

Ebb-Real:
Just digitigrade shoes overall, is it worth maintaining a feature that barely effects gameplay for only a few people that make an optional choice to have digitigrade legs. The shoes could easily be broken by TGUs in the future.

Ebb-Real:
They also would need a sprite rework and maintenance considering people criticizing that a lot.

SpaceDragon00:
If I removed every feature I couldn't be bothered to fix there would be nothing left. The first instinct when you have a bug shouldn't be a full feature removal. Comment out the species and vending machine code as a testmerge if you want a quick and easy way to avoid the bug until someone fixes it properly

> Just digitigrade shoes overall, is it worth maintaining a feature that barely effects gameplay for only a few people that make an optional choice to have digitigrade legs

If it's too much work to code then don't code. Coding is voluntary. Don't want to fix it? Don't, and wait until someone else does

Also, if you aren't coding for the sake of the players then who are you coding for? This just sound like "I don't like or use this so remove it."

Ebb-Real:
yeah, you're right this may not be the right way to do it now I know its possible to fix.

SpaceDragon00:
Actually, let me rephrase this in a way that is less belligerent and more succinct:

If you remove the bug then the bug will never be fixed because no one has any way of knowing the bug was ever there to begin with

@Ebb-Real

Ebb-Real:
> Actually, let me rephrase this in a way that is less belligerent and more succinct:
> 
> If you remove the bug then the bug will never be fixed because no one has any way of knowing the bug was ever there to begin with
> 
> @Ebb-Real

I get that, thanks for letting me know that it is fixable in a non-insane way anyhow.

OneAsianTortoise:
Can't we just remove lizard

JohnFulpWillard:
Currently the new limbs system has fucked over support for digitigrade shoes. They tried adding their own, in a system where shoes will automatically become 'digitigrade' ones if it has the right sprites for them. We don't do this, and because of it, it is WAY harder to keep support for ours.

> Don't want to fix it? Don't, and wait until someone else does

That's the problem. No one will. It's been over a month with this bug and not one person shows interest in fixing it.
This shit breaks once every 2 TGU's and it's been very annoying to maintain it. New limbs was just the final nail in the coffin for me.

In general, one of three things will happen:

1. People will see the removal PR and say 'wait we can just fix it instead'
2. People will say after the removal 'can I just readd it if I fix it'
3. Complain that it is removed but do nothing about it

No one will fix the bug while it is still a bug. Cloning, Clock cult, Nanites, Old circuits, and many other examples can be cited for this. Even on the Fulp repo, no one wanted to fix the problems with Brig Physicians, but all complained after its removal that it should've just been 'fixed' instead, but didn't want to be the one to do it. 

You constantly act like any removal is bad and it should be fixed instead. Nanites was removed instead of 'fixed', and it turned out to be a good change. Removals aren't bad, so stop acting like it's the end of the world when any feature is removed.

> In short, if it's too much effort to fix the bug, why isn't it too much effort to try and remove all the content?

If it was good content then it would be on TG. We had it on Fulp because it was grandfathered from the old fulp repo, and you should be well-aware the type of things that were accepted then.

> The first instinct when you have a bug shouldn't be a full feature removal.

It is not first instinct. It's been something I've wanted to do for a long time. Again, if it was a good idea, it would be on TG.


And speaking of all of this, Fulp code in general feels like its been bloated a bit overtime, I'd like to go over it and kill a few things one day.

Ebb-Real:
digis would be different from normal lizards like at TG and also if the no hardstun thing gets added at TG it would make sense at fulp

FlamingEmu595:
if you dont like digi shoes that much why on earth give people the option then just remove fuckin digi limbs if it's that big a deal

Ebb-Real:
Digi shoes were an old feature on fulp that was just kinda left in the dirt after being implemented only maintained by the people that would have to change them when updating with TG. If they should exist it should be on the upstream and not here both because they have a lot more contribs and also you woulden't have to worry about making it modular. But iirc digi shoes were attempted at TG but denied. I don't see the point in headbutting TGs design choices and causing bugs for a sliver of another sliver of players.

Removing digis entirely would be the same thing working against TGs code instead of doing it directly on TG.

PR #642, opened 2022-05-18
Bloodsucker throne properly works now

FernandoJ8:
## About The Pull Request

Should fix the bug where the bloodsucker throne doesn't actually work because it tries to send the message to the antagonist datum instead of the vassal's mob

## Changelog

:cl:
fix: bloodsucker bloodthrone now works as intended
/:cl:

### DIFF
--- a/fulp_modules/features/antagonists/bloodsuckers/code/bloodsucker/bloodsucker_crypt.dm
+++ b/fulp_modules/features/antagonists/bloodsuckers/code/bloodsucker/bloodsucker_crypt.dm
@@ -740,7 +740,9 @@
 	var/rendered = span_cultlarge("<b>[user.real_name]:</b> [message]")
 	user.log_talk(message, LOG_SAY, tag=ROLE_BLOODSUCKER)
 	var/datum/antagonist/bloodsucker/bloodsuckerdatum = user.mind.has_antag_datum(/datum/antagonist/bloodsucker)
-	to_chat(bloodsuckerdatum.vassals, rendered)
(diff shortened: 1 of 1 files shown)


fulp_modules/features/antagonists/bloodsuckers/code/bloodsucker/bloodsucker_crypt.dm
```diff
@@ -740,7 +740,9 @@
 	var/rendered = span_cultlarge("<b>[user.real_name]:</b> [message]")
 	user.log_talk(message, LOG_SAY, tag=ROLE_BLOODSUCKER)
 	var/datum/antagonist/bloodsucker/bloodsuckerdatum = user.mind.has_antag_datum(/datum/antagonist/bloodsucker)
-	to_chat(bloodsuckerdatum.vassals, rendered)
+	for(var/datum/antagonist/vassal/receiver in bloodsuckerdatum.vassals)
+		var/mob/receiver_mob = receiver.owner?.current
+		to_chat(receiver_mob, rendered)
```
JohnFulpWillard, line 745:
```suggestion
	for(var/datum/antagonist/vassal/receiver as anything in bloodsuckerdatum.vassals)
		to_chat(receiver.owner.current, rendered)
```
we already know they have a mind and a body because otherwise they wouldnt be in the vassals lise
We also know that only vassal antag datums will be in the bloodsucker's vassals list, so we can as anything it.

FernandoJ8:
What the fuck that's the first time I ever see the icon diff bot fail

PR #667, opened 2022-06-21
Fixes our CI, Putting on cat ears gives just a bit more brain damage

SgtHunk:
## About The Pull Request

CI error was caused by felinids spawning with the aphasia brain trauma which when qdeleted runtimes lol
Also the kitty ears give you 100 instead of 75 brain damage now i think it improves the game a bit

## Changelog

:cl:
sounddel: Ci doesn't fail anymore
/:cl:


JohnFulpWillard:
What sucks is that I can’t replicate this by unit testing giving a brain trauma then immediately running on life and taking it away

SgtHunk:
its because the brain gets qdeleted and stuff on aphasia doesn't get fully cleared idk

PR #670, opened 2022-06-22
removes deputies

JohnFulpWillard:
## About The Pull Request

removes deputies in its entirety, and removes basically all fulp unique job code, since we no longer have any fulp jobs now.

## Why It's Good For The Game

I'm too tired to explain why. TG gave departmental officers AA to their department so deputies lost their niche, now people who play deputies can just play sec off.

I kept the deputy sash in because it was in a separate module, in case deputies are ever re-added, and the mutadone medipen is still sticking around as it's used in sec's cargo console.


Pepsilawn:
Fixes #673

PR #672, opened 2022-06-24
Saves our parallel map testing tests?

Pepsilawn:
## About The Pull Request

The dummy mob used for screenshot tests is undefeatable we should just bow to them and stop trying to give them traumas as we have no way of winning against it with the way its assigned its race


SgtHunk:
^ full of /mob/living/carbon/human/dummy/consistent

PR #693, opened 2022-06-27
Comments out screenshot tests

Pepsilawn:
## About The Pull Request

Undefeatable until tg provides an alternative way of disabling it but by editing its files.


Pepsilawn:
Only fulp pr that passes checks ever

PR #701, opened 2022-06-27
Separates Heliostation's  Aft Surgery into two separate rooms.

maxspells:
## About The Pull Request

helio surgery
## Why It's Good For The Game
I really love having separate rooms for surgeries. The smartfridge is in the center of the room so it can be used by either side. Each room has shutters that can can be closed for privacy. I moved the wall into the security room a bit so I had to slightly redesign it and remove the chairs, but other than that everything is mostly unchanged. Playtested it and everything seemed to work fine, but this is my first map PR so I'm sure there's probably something I missed.

### DIFF
--- a/_maps/map_files/Heliostation/Heliostation.dmm
+++ b/_maps/map_files/Heliostation/Heliostation.dmm
@@ -61544,7 +61544,7 @@
 /obj/machinery/atmospherics/pipe/smart/simple/supply/hidden/layer4,
 /obj/machinery/door/firedoor,
 /obj/machinery/door/airlock/medical{
-	name = "Surgery"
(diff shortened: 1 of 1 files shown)


Pepsilawn:
Surgerya requires access but not surgeryb?

Pepsilawn:
You also didnt edit your cameras

maxspells:
Thanks, looking into fixing those now!

Pepsilawn:
Camera name doesn't matter what you're looking for is the c_tag, also needs to be added to medbay's network

Pepsilawn:
Why are we using blast doors instead of shutters

PR #727, opened 2022-07-12
Nerfs SeleneStation's medbay maints' crate

Pepsilawn:
Some joker put a x8 maintenance loot spawner over a crate spawner which can spawn as a mini fridge that has a low storage capacity & that makes it all shit itself

image


SgtHunk:
The joker with a x8 maintenance loot spawner:

PR #728, opened 2022-07-13
Saves Selene's brig

OperativeLyn:
## About The Pull Request
What the hell 🤯 
image


SgtHunk:
y is it not using the directional subtype 😨

OperativeLyn:
> y is it not using the directional subtype 😨

I was gonna do it but none of the cameras use it so I'd have to change every other camera to be consistent

PR #744, opened 2022-07-20
Fixes beefman spawners

JohnFulpWillard:
## About The Pull Request

Because spawners use randomize_human, which is a hardcoded base proc not attached to anything and cannot be modularized, we are instead setting it to randomize on the species' gain.

image

Sucks, but works.


SgtHunk:
randomize_human() sucks

PR #753, opened 2022-07-28
Redoes all of Helio's disposals

Pepsilawn:
Idk if it'll even compile this was guillaume's idea


Pepsilawn:
Nothing went as expected this game sucks

PR #776, opened 2022-08-06
Removes our TG edit from basemap

JohnFulpWillard:
## About The Pull Request

I'm trying to get our TG edits back down to below 10, which is quite hard because what we used TG edits for has shifted overtime from 'necessary for our features' to 'necessary for our codebase' for hardcoded stuff like TGUI, screenshot tests, and GAGS.

This is at least removing an easy one that I found out worked like, 5 minutes ago, so I'm doing it.

## Why It's Good For The Game

TGU is a little harder to accidentally screw up, and remove all our maps.


fulp_modules/mapping/_basemap.dm
```diff
@@ -0,0 +1,8 @@
+#ifndef LOWMEMORYMODE
+	#ifdef ALL_MAPS
+		#include "..\..\_maps\map_files\PubbyStation\PubbyStation.dmm"
+		#include "..\..\maps\map_files\HelioStation\HelioStation.dmm"
```
SgtHunk, line 4:
```suggestion
		#include "..\..\_maps\map_files\HelioStation\HelioStation.dmm"
```

SgtHunk (changes requested):
💀

PR #777, opened 2022-08-08
Lets mentors use mentor manager

JohnFulpWillard:
## About The Pull Request

Mentors lied to me and told me this works, when it doesn't
Now mentors can use mentor manager, though they will still get alerts of not having perms. they can deal with it.


JohnFulpWillard:
hey 777

OperativeLyn:
the lucky number...

PR #788, opened 2022-08-10
mime emote update

567Turtle:
## About The Pull Request

As a mime main, I always hate when I go walkin' around, get a virus and start coughing audibly, I'm a mime for farts sake! So I went in, I moved a few things around and I jerry-rigged a fix for it

## Why It's Good For The Game

View above, as a mime main the urge to go on a violent rampage when I *sneeze uncontrollably and _audibly_, a mime should never be forced to break his vow in such a way. I think I got most of the emotes that imply making a noise with your mouth, comment or whatever if you think of one I forgot.

## Changelog

:cl:
add: added more unique mime emotes
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/emote.dm
+++ b/code/modules/mob/living/carbon/human/emote.dm
@@ -5,6 +5,7 @@
 	key = "cry"
 	key_third_person = "cries"
 	message = "cries."
+	message_mime = "sobs silently."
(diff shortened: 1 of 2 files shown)


FernandoJ8:
This should probably go on tg and not fulp

567Turtle:
oh yeah I forgot fulp did that auto update thing, sure I'll go do that instead

PR #815, opened 2022-09-10
Removes an undocument balance change

Pepsilawn:
## About The Pull Request

What is says on the tin

## Why It's Good For The Game

Doesnt feel right to keep around when players wouldn't be aware of it

### DIFF
--- a/fulp_modules/Z_edits/heart.dm
+++ b/fulp_modules/Z_edits/heart.dm
@@ -1,3 +1,0 @@
-/datum/action/item_action/organ_action/cursed_heart
-	// You can manually beat it without having to be awake.
-	check_flags = NONE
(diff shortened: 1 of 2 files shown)


SmoSmoSmoSmok:
nice checks kid

SgtHunk:
>Doesnt feel right to keep around when players wouldn't be aware of it

We can put it on the wiki in that case.
If this balance change is removed, then this will be a noob trap for bloodsuckers.
I'd much rather this simply be documented on the wiki than removed.

Pepsilawn:
> > Doesnt feel right to keep around when players wouldn't be aware of it
> 
> We can put it on the wiki in that case. If this balance change is removed, then this will be a noob trap for bloodsuckers. I'd much rather this simply be documented on the wiki than removed.

The item is noob bait to begin with you lose so much blood if you lag out/go afk/don't click the button on your screen every 3 seconds i just dont think it makes sense, also nerfs the corazargh chem which is why it was denied on tg to begin with

PR #822, opened 2022-09-25
Adds layer adapters to heliostation atmos gas chamber outlets

Taiteilija:
## About The Pull Request

Adds layer adapters to Heliostation atmos outputs.

## Why It's Good For The Game

QOL change that is standard on tgstation based maps. Removes the need for atmos techs to do this at shift start so a single output can be used for multiple purposes.

## Changelog

:cl:zhepyrlion
qol: heliostation atmospherics gas chamber outputs now have layer adapters
/:cl:


OperativeLyn:
Why only Helio and not the rest of the Fulp maps?

PR #848, opened 2022-11-09
Fixes cruelty squad gloves

Sheits:
The gloves on the suit were broken so this should fix it maybe possibly


Sheits:
umm epic fail

PR #854, opened 2022-12-04
Techweb-purchase able programs start locked and behind Command access

JohnFulpWillard:
## About The Pull Request

Moves researching back to a Command job by locking it behind Command access, setting all consoles/programs to be locked by default.

Emagging still unlocks the RD console.

## Why It's Good For The Game

Command tries to cooperate and make sure research goes fine for everyone, but that's hard to do when everyone in science can research anything they want, whenever they want. Instant gratification isn't always good, especially when it takes away the ability for command to manage their department.

Fixes situations like this
image

## Changelog

:cl:
balance: Techweb-purchasing apps now start locked and require Command access (or an emag) to unlock.
/:cl:

### DIFF
--- a/fulp_modules/Z_edits/science_edit.dm
+++ b/fulp_modules/Z_edits/science_edit.dm
@@ -1,0 +1,13 @@
+/obj/machinery/computer/rdconsole
+	//sets lock access to command
+	req_access = list(ACCESS_COMMAND)
+	//starts it off locked
+	locked = TRUE
+
+/datum/computer_file/program/science
+	//starts it off locked
+	locked = TRUE
+	//sets lock/download access to command
+	lock_access = ACCESS_COMMAND
+	required_access = list(ACCESS_COMMAND)
+	transfer_access = list(ACCESS_COMMAND)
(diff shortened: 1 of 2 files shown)


KittyNoodle:
can we please not go back to this

JohnFulpWillard:
It was reverted on TG because:
1- They generally do not have heads of staff (or any wiling to research, or become acting RD, or the RD is playing it just in case they roll antag), which is generally not a problem on Fulp who has antag command disabled and holds command to higher standards
2- It was easily circumvented through just building an R&D board that the circuit imprinter can make roundstart, which is not a problem here now

> consoles accessible is easier for both players to understand

How so?

> and for moderators to moderate.

How so? Consoles are all locked by default so the only way someone would be able to research is through direct approval from someone who has access, which I don't see as anything that requires moderator approval.

remove32:
from a policy standpoint i think it is better to have the freedom to do research should RD be busy or otherwise indisposed. there can be good RP to get from head/research staff interactions concerning who does what research. i think its better to use the current system for its flexibility. to my knowledge, we haven't had significant issues with this approach. 

all in all, i want to rely more on the players to get RP out of it and report it to admins when the need arises. player interactions and flexibility is my priority

PR #855, opened 2022-12-04
Promotes the QM to command

Pepsilawn:
## About The Pull Request

Apparently QM becoming a head happened like 6 months ago in June which is like half a year ago so this line isn't very true anymore


zzzmike:
wouldn't it be better to get qm added to DEPARTMENT_BITFLAG_COMMAND rather than add an exception like this?

Pepsilawn:
would it now

JohnFulpWillard (approved):

zzzmike:
> would it now
oh that's what you did, removed the exception. I read it backwards, woops haha

zzzmike (approved):

Issue #847, opened 2022-11-07
Heliostation Law office's fax machine is public


JohnFulpWillard:
The fax machine on Heliostation's law office is at the entrance, meaning anyone can walk in and use it to impersonate a Lawyer.

Issue #845, opened 2022-11-02
Heliostation CMO PDA painter is inaccessible


JohnFulpWillard:
## Reproduction:

![image](https://user-images.githubusercontent.com/53777086/199420539-ef30f14c-3fb4-4301-9431-90f7b291e841.png)

PR #859, opened 2022-12-06
Fixes Helio Fax Machine Placement Problems

Tramzz:
Fixes Heliostation's Law office fax machine being public, and fixes CMO's PDA painter from being inaccessible due to the fax machine's placement.
## About The Pull Request
Addresses issues #847 and #845 on Helio by moving the placements of their fax machines, and adding an extra fax machine to the law office.

Goes from this:
image
To this:
image

And here's the law office:
image
## Why It's Good For The Game
If the CMO doesn't have access to their PDA painter how else will they demote all the Walter White's that blow up chemistry for the 80th time? On a serious note, without access to the PDA painter, CMO can't do PDA changes within their own department, so I'd argue its fairly important. Also lawyers can no longer be impersonated using their fax, since their fax is no longer open to the public.
## Changelog
:cl:
fix: fixed CMO's PDA painter being inaccessible
fix: fixed the law office's fax machine being public
add: added two machines to the individual law offices
/:cl:

### DIFF
--- a/_maps/map_files/Heliostation/Heliostation.dmm
+++ b/_maps/map_files/Heliostation/Heliostation.dmm
@@ -395,14 +395,14 @@
 /area/station/security/prison)
 "abe" = (
 /obj/machinery/firealarm/directional/east,
-/obj/structure/bed/dogbed/runtime,
-/obj/item/toy/cattoy,
 /obj/effect/turf_decal/tile/blue/anticorner/contrasted{
 	dir = 4
 	},
 /obj/structure/disposalpipe/segment{
 	dir = 9
 	},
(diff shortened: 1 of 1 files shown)


Pepsilawn:
Having two fax machines with the same name sounds annoying when the list of faxable machines will list both with no difference

Tramzz:
> Having two fax machines with the same name sounds annoying when the list of faxable machines will list both with no difference

That's a good point, I can probably see if one can be renamed or something like that so they don't share the same title. Would North and South sound fine considering the two offices?

OneAsianTortoise:
No way, actual changelog in fulp repo PR

SgtHunk:
good pr

PR #861, opened 2022-12-06
Adds support for non-range checking powers

JohnFulpWillard:
## About The Pull Request

Adds in support for powers that don't need to range check whatever they're hitting.

## Why It's Good For The Game

Some powers let you click wherever you want and don't need a range check, and before this was done by setting it to 99, which I think is just absurd. Lunge also set it to how many tiles a person can see, however, widescreen broke that because suddenly people saw more than just 6 tiles away. This aims to fix that permanently by just removing both those cases and just simply not checking distance if it isn't necessary.

This will also be useful for https://github.com/fulpstation/fulpstation/pull/860


ShadowAKT:
Good long-term fix for the issue I brought up.

PR #862, opened 2022-12-10
Sol is now a Subsystem, Coffins lock themselves, Bloodsuckers don't constantly die, probably

JohnFulpWillard:
## About The Pull Request

Turns Sol into a subsystem and hooks Bloodsuckers onto it via signals instead of doing a ton of for() loops anywhere. This made Sol incredibly fucking fast, so I halved the speed so it only ticks every 2 seconds.
I also gave Sol a 90 second randomness every night, so night 1 is 10 minutes, second night can be 8-12, then 6-14, but it's random so it doesn't only increase/decrease by 2, can also just increase/decrease by 1 or not at all. This is just meant to prevent people's ability to metagame.

I also improved the sunlight hud to update with regular bloodsucker updates to avoid some useless proc overhead and fixed Coffins not locking by themselves.

These changes also seem to have fixed the issue of a clan-less Bloodsucker dying constantly. I've tested while both inside and outside of a Clan and it seems to work fine now, I dunno exactly what the cause was, but as I can't replicate it, I just have to hope this fixes it.

Finally, I removed Monster hunter's Sol warnings, I didn't find them adding anything beneficial to the game and generally was just a single message saying "Bloodsuckers are sleeping lol". I had added this and now wish to remove it because I think the game is better without it.

## Why It's Good For The Game

Fixes and addresses many concerns and makes Sol more resilient to breaking, giving it consistency in a round and making it less metagame-able.

## Changelog

:cl:
add: Sol is now a subsystem and its timers should now be consistent.
fix: Bloodsuckers should no longer die from sleeping through Torpor/Sol.
fix: Bloodsucker HUDs all update together.
fix: Bloodsuckers now automatically lock their Coffin when entering it.
balance: Monster Hunters no longer have a message to tell them Bloodsuckers went to bed.
balance: Sol now has slightly randomized times between nights.
/:cl:


OneAsianTortoise:
@StoneMason  L bozo

SgtHunk:
no one has complained about this in the 3 days it was tmd so

PR #863, opened 2022-12-11
Helio White ship

Horatio22:
## About The Pull Request

A white ship for helio. It's a pirate drug den.  It (now) looks like this:

image

Now with meth-making hints!

Big thanks to Ruvic for the help with the meth setup!

## Why It's Good For The Game

The /tg/ maps all have their own white ships, so helio should get its own too.

## Changelog
:cl:
add: Added a white ship for helio
/:cl:

### DIFF
--- a/_maps/heliostation.json
+++ b/_maps/heliostation.json
@@ -6,7 +6,7 @@
 	"shuttles": {
 		"cargo": "cargo_helio",
 		"ferry": "ferry_fancy",
-		"whiteship": "whiteship_meta",
+		"whiteship": "whiteship_helio",
 		"emergency": "emergency_helio"
 	},
 	"job_changes": {
--- a/_maps/shuttles/whiteship_helio.dmm
+++ b/_maps/shuttles/whiteship_helio.dmm
@@ -121,10 +121,6 @@
 "jV" = (
 /turf/closed/wall/mineral/titanium,
 /area/shuttle/abandoned/cargo)
-"kp" = (
-/obj/effect/spawner/structure/window/reinforced,
-/turf/closed/wall/mineral/titanium,
-/area/shuttle/abandoned/medbay)
 "kq" = (
 /obj/machinery/chem_heater,
 /turf/open/floor/engine,
@@ -741,7 +737,7 @@ Oo
 Oo
 fk
 yQ
-kp
+dh
 is
 rJ
 rJ
--- a/_maps/shuttles/whiteship_helio.dmm
+++ b/_maps/shuttles/whiteship_helio.dmm
@@ -686,8 +686,8 @@
 /turf/open/floor/wood,
 /area/shuttle/abandoned/engine)
 "Gi" = (
-/obj/machinery/door/airlock/alarmlock,
 /obj/structure/cable,
+/obj/machinery/door/airlock/public/glass,
 /turf/open/floor/oldshuttle,
 /area/shuttle/abandoned/bar)
 "Gp" = (
--- a/_maps/shuttles/whiteship_helio.dmm
+++ b/_maps/shuttles/whiteship_helio.dmm
@@ -122,8 +122,7 @@
 /turf/closed/wall/mineral/titanium,
 /area/shuttle/abandoned/cargo)
 "kp" = (
-/obj/effect/spawner/structure/window/reinforced/tinted,
-/turf/closed/wall/mineral/titanium/nodiagonal,
+/turf/open/floor/engine/air,
 /area/shuttle/abandoned/medbay)
 "kq" = (
 /obj/machinery/chem_heater,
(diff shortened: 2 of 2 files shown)


JohnFulpWillard:
What is the whiteship's theme? I assumed it was 'medical' from how you have an entire medical techfab and plumbing room, but you also have a bar and botany

With the space cat, pirates, RCD, and lack of all chemistry equipment, this looks like it's trying to mix in a bunch of other ruins together.

SgtHunk (changes requested):
As mentioned above, this ship feels a bit all over the place and could use a more streamlined idea of what it should be.
But your idea was meant to be a chemistry plumbing ship iirc, so it should focus on that aspect.
It's not a bad concept but it could use a lot of polishing.
I'd check the areas in strongdmm to see if they 100% match.

The rooms in general look bland and could have a more atmospheric look to fit with the ambience, the entrance has too much lighting and looks quite big, it feels empty to stare at.
The main room has things all over the place, which although isn't bad there is no indicator to where they should be or how everything interacts. Things are in the corners of the room to fill in space, and although it's a very big space everything just feels shoved in like there wasn't anywhere else to place them.
Hydroponics and the Bedrooms look off due to the awkward diagonal movement that'll occur in-game because of the shuttle's shape, covered with tinted windows that serve no real purpose? There is no reason for Hydroponics to be covered in tinted windows, or for both sides to be swarmed in windows in general.
The Bar's trash can being one tile deeper feels wrong, and it's covered in windows for no apparent reason too. The normal and reinforced windows clash in design with no wall to separate them.
The Bathroom's objects aren't in any wall and need to be fixed.
The Cockpit has a lot of lights and they're a tile deeper to fill in blank space, same design issues. It uses tinted windows?? why would you not want to see what's up ahead in the cockpit.

Tl;dr, a lot of design issues that can be fixed.

1- Decals. Decals are IMPORTANT for this. Indications of where things should be, adds a lot of value to the atmosphere of the place in general.
2- Lighting. Same as above, Don't use too much lighting or else it'll look too sterile as an environment - this is meant to be a decaying shuttle raided by a pirate crew. Make it seem like so. Broken lights, light fixtures with nothing in them, cables spread around, etc.
3- Tiles. Adding to 1, You keep using the same tiles to fill an entire room and it just looks *bad*. An airlock has no reason to have reinforced tiles, neither does Hydroponics. Carpets ideally should only be rectangular, with fitting tiles in each room to demonstrate its coziness and risk associated with it.
4- Shape. Diagonal shuttle doesn't sound too bad in concept until you realize the issues that come with it, In its current shape a lot of the space is wasted and there's a lot of diagonal movement on the sides, with objects being filled in on those small spaces, it just doesn't give a very clean or effective look.

SgtHunk:
oh and also the footprints are backwards lol

SgtHunk (changes requested):
Looks better now, but could still use some design changes.

There being a main crosswalk running through all areas is a nice idea, but in this case I believe it'd look better if each room served independent of the rest, to allow for more creativity in its design that is unbothered by a specific tile running through the middle of them.

1- Add more objects to make the rooms feel more alive, you can use the pirate ships for reference, mainly the og one, since it has a similar structure.
2- Floor lights aren't really necessary seeing as the rooms aren't big enough to need them.
3- In general, check the positioning of the structures and items. The dorms need to feel welcoming as dormitories. The chemistry room feels hard to navigate because of this, moving the chemistry part to one side and the items required to another would clear this, freeing up the middle and dividing the areas (which also would look good decal'd). A completed chem room will look nicer if done that way.
4- Check the MDB render, window showing nothing and missing wall. Break room has 2 air alarms.  There are still 2 shower heads. Also make sure all windows have plating underneath and not a regular tile.

Addust:
are you sure the pacman and SMES are set up correctly?

Addust:
also please just use VV'd titanium windows for the tinted windows if those haven't been removed for whatever reason, because aesthetically it's more pleasing

Horatio22:
> also please just use VV'd titanium windows for the tinted windows if those haven't been removed for whatever reason, because aesthetically it's more pleasing

No.

SgtHunk (changes requested):
Slowly making progress.

1- Check MDB, missing wall between cockpit access and bar. Will look a bit better if the 3rd tinted window is removed, since it's connected to the south wall.
2- Suggestion from previous review remains regarding the main chemistry room, seeing as a built chem setup will make moving through the room difficult and awkward. And ofc decaling too.
3- The footprints are still backwards (?)
4- Dorms carpet is weirdly shaped, also MDB shows it looking weird too with even a missing wall, dunno what happened there.

SgtHunk:
window is useless and should be a wall instead

68747470733a2f2f6d6462322e616666656374656461726330372e636f2e756b2f696d616765732f3331303139313836312f313035363534363036383332f612f302f302d61646465642e706e67

this wall is invisible !!! (should be non-diag)

68747470733a2f2f6d6462322e616666656374656461726330372e636f2e756b2f696d616765732f3331303139313836312f31303536353436303638332f612f302f302d61646465642e706e672

istg its the last i'll nitpick

PR #890, opened 2023-01-29
Bloodsuckers: Refactors, QoL, Fixes

JohnFulpWillard:
## About The Pull Request

- Fixes the Lunge ability, you now only spin when you're charging up at someone.
- Lunge also is limited to 6 tiles when it's upgraded to not spin anymore.
- I also tried to de-hardcode lunge's procs and tried to cut down on the overuse of pathing target/user to carbon on each individual one.
- Removes Vampire teams, it now checks if there's any Bloodsucker antag datums to turn Sol on/off, since it's now a subsystem.
- Removes 'frenzy_blood_drank' and 'frenzies' vars and the objectives that required them, because they were unused.
- Adds Tzimisce to the Curator's book, also de-hardcodes it and improves the UI significantly. Now instead of opening a book panel, all bloodsucker stuff is in the UI directly, so you don't need to constantly be opening the book.
- Removes "Hunter purchaseable" actions, since it wasn't used at all.
- Revenge Vassals now only show up in the roundend report independently if their Bloodsucker master has died.

## UI examples

Makes the Antag page consistent with the Antag tip one for UI style
image

Kindred archives also use it as it's now the "Bloodsucker" style, also showing the nice UI it now has
It's also been refactored in the backend, built up on tsx instead of js.
image

## Lunge demonstration

https://user-images.githubusercontent.com/53777086/215317919-88457983-3181-4013-a7b4-36dad6267ec1.mp4

## Why It's Good For The Game

Swag fixes & Better UI & Awesome de-hardcoding & Removes cringe vampire clan that confuses people reading the code & More flavortext

## Changelog

:cl:
add: Archives of the Kindred now have a better UI that works better and without the need to constantly open and close white-mode pure HTML pages.
code: Archives of the Kindred no longer has hardcoded clans and instead use all clan subtypes in their book.
refactor: Bloodsuckers no longer use a 'vampire clan' team (which wasn't visible to the player anyways).
fix: Lunge now works
fix: Lunge now respects range (if upgraded enough, it's now a feature lol)
fix: Bloodsuckers' Antag information now has a nice color and fits the entire page without the need to scroll.
/:cl:


fulp_modules/features/antagonists/bloodsuckers/code/powers/fortitude.dm
```diff
@@ -82,7 +82,7 @@
 /datum/action/bloodsucker/fortitude/hunter
 	name = "Flow"
 	desc = "Use the arts to Flow, giving shove and stun immunity, as well as brute, burn, dismember and pierce resistance. You cannot run while this is active."
-	purchase_flags = HUNTER_CAN_BUY
+	purchase_flags = NONE
```
remove32, line 85: was this proc ever used? does the monster hunter not use this?

fulp_modules/features/antagonists/bloodsuckers/code/powers/targeted/lunge.dm
```diff
@@ -3,20 +3,27 @@
...
-		There is an exception to this, those wearing Riot gear, and Monster Hunters, will be passively grabbed instead.\n\
+		Riot gear and Monster Hunters are protected and will only be passively grabbed.\n\
 		You cannot use the Power if you are already grabbing someone, or are being grabbed.\n\
+		If you grab from behind, or from darkness (Cloak of Darkness works), you will knock the target down.\n\
 		If used on a dead body, will tear their heart out.\n\
 		Higher levels increase the knockdown dealt to enemies.\n\
-		At level 4, if you grab from behind or from darkness (Cloak of Darkness works), you will knock the target down."
-	power_flags = BP_AM_TOGGLE
+		At level 4, you will no longer spin, but you will be limited to tackling from only 6 tiles away."
+	power_flags = NONE
```
remove32, line 14: how does changing this flag alter its function?

fulp_modules/features/antagonists/bloodsuckers/code/powers/fortitude.dm
JohnFulpWillard: Monster hunters have the action granted to them, they never purchase powers so this did nothing.

fulp_modules/features/antagonists/bloodsuckers/code/powers/targeted/lunge.dm
JohnFulpWillard: The flag causes it to start processing as you click on the button, which starts the spinning. Instead we start processing on the Lunge itself when they click on a target, so they spin when they are actually charging up at someone.

remove32: feels weird defining things as NONE like this but hey, if it works...

PR #904, opened 2023-02-02
Puts Heliostation to TG Mapping standards

SgtHunk:
## About The Pull Request

Removes varedited dir pipes from Heliostation (for the most part), replacing them with the tg standard piping.
These normal pipes still exist in Atmospherics for easier visibility in the mapping editor without having to hide them with layers filter, but everything else is the manifold 4-way.

## Why It's Good For The Game

Our maps should follow standards, keeping antiquated piping does nothing but make work harder for contributors who have to deal with it every time they have to make a change. The visual element is only in the mapping editor, and as mentioned above can simply be hidden if it annoys the mapper.

## Changelog

:cl:
add: Smartpipes to Helio
/:cl:


FernandoJ8:
The hero we needed. The hero we deserved.
Borzoi

PR #906, opened 2023-02-02
Fixes the position of the grav gens on fulp maps

FernandoJ8:
## About The Pull Request

They were offset one tile to the right due to our second-to-last TGU. This PR fixes that.
## Changelog
:cl:
fix: fixed the position of grav generators on Heliostation, Selenestation and Pubbystation
/:cl:


FernandoJ8:
> tg issue, if this is merged we'll need to revert this pr again the tgu after

since we'll have to wait until march tgu it seems necessary

Pepsilawn:
This creates another issue down the line that we will need to keep in mind for a month from now which will probably be forgotten about where our grav gens will be offset a tile to the left

Pepsilawn:
Or just adds the correct offset in code so itll get overwritten by tgu instead of having to fix all maps twice for the same reason

PR #914, opened 2023-02-17
February fixes batch (Abductors, cameras, grav gen, rpg titles, timestop)

SgtHunk:
## About The Pull Request

Since Pepsilawn is STALLING Traditional Games Update I am forced to make this pull request.

- Fixes abductors probably
- Fixes paradox clone status effects breaking it
- Fixes gravity gen being off-center and not giving the aura
- Makes RPG titles not obnoxious with runechat
- Fixes timestop being smaller than it should be

## Why It's Good For The Game

fix


SgtHunk:
TGU happened

PR #938, opened 2023-03-23
fixes a runtime

tattax:
RequireDisloyalty runs then checks AmValidAntag, which itself alone doesn't have a mind check, could use conditionals


eucalyptusJ (approved):

fulp_modules/features/antagonists/bloodsuckers/code/bloodsucker/bloodsucker_conversion.dm
```diff
@@ -5,6 +5,9 @@
  * target - The person we check for antag datums.
  */
 /datum/antagonist/bloodsucker/proc/AmValidAntag(mob/target)
+	if(!target.mind)
+		return
+
 	if(target.mind.unconvertable)
```
JohnFulpWillard, line 11:
```suggestion
	if(!target.mind || target.mind.unconvertable)
```

Banned just means they won't prompt them to accept disloyalty

JohnFulpWillard (approved):

PR #958, opened 2023-04-06
replaces a set of codersprites with another set of codersprites

SmoSmoSmoSmok:
## About The Pull Request
before:
before

after:
after


SmoSmoSmoSmok:
also fixes issue in infils introduced by the tgu

PR #975, opened 2023-04-19
Removes 2 TG edits & TGU

JohnFulpWillard:
## About The Pull Request

Deletes our yarn thing
Moves the fulp TGUI overwrite to an edit since we already edited the file.
We shouldn't be writing our own dependencies, we should instead be using TG's because they keep it up-to-date.

Does a TGU, removing our tg edit from GAGS greyscale config

## Why It's Good For The Game

Removing 2 tg edits and making it easier to maintain the code in the future.
I had to do TGU in this PR because the UI was updated on TG.


JohnFulpWillard:
i deleted selene

PR #976, opened 2023-04-23
Removes the Fulp medipen re-define

JohnFulpWillard:
## About The Pull Request

TG added a mutadone medipen with the exact same path as ours, rendering ours useless.
Theirs also has an extra 5 units in it.


remove32:
sneaking in some code quality updates while you're at it eh?

JohnFulpWillard:
Yea just swapping string to define, I did it in the past but I think someone accidentally reverted

PR #977, opened 2023-04-23
GET BEEFY GET BLOODY

JohnFulpWillard:
## About The Pull Request

Beefmen now only bleed out at 25C (used to be 24C)
Beefmen regenerate blood when they aren't bleeding (This was a feature on the old fulp repo but stuck around as unused code here)
Beefmen no longer get blood from eating
Beefmen will always bleed over people even if they aren't bleeding
Other people can now once again attach limbs to Beefmen for them, and now it works with their tongue too.
Beefmen start taking burn damage the moment they reach the 25C needed to bleed out, however they will only take HALF the burn damage. This makes them better at resisting fire damage short-term, at the cost of more bleeding long-term.

## Why It's Good For The Game

GET BEEFY GET BLOODY GET MEATY

Beefmen is meant to be balanced around constantly shifting blood volumes and limbs flying around, this will hopefully make this happen more organically without requiring code-diving knowledge while not being as punishing to new players.


OneAsianTortoise:
@SmoSmoSmoSmok, how much u pay this dude

PR #984, opened 2023-05-05
Adds HoochMaster Pubbystation's Bar

ackbarscrowbars:
## About The Pull Request

**Adds a HoochMaster Deluxe** (Bar variant of CondiMaster/ChemMaster) **to the Pubbystation Bar**, adds a 2nd Bar Requests Console in Bar Storage (Next to the double-barreled shotgun), moves around most of the wall-mounted stuff in the Bar (Newscaster, Request Console, Intercom, Fire Alarm), and adjusts the Light switch's X/Y position a little. No changes to the Lounge.

## Why It's Good For The Game

This tweaks the layout of the wall-mounted objects in a mildly more appealing manner, as well as adding a HoochMaster Deluxe, which Pubby's bar **doesn't have**. This forces the Bartender(s) to take or build a ChemMaster or equivalent machine if they want to make cocktails of any complexity. I also believe the storage room could use an additional request console. If nothing else, it's a little extra clutter.

## Changelog

:cl:
add: Adds a HoochMaster Deluxe (Bar variant of CondiMaster/ChemMaster) to the Pubbystation Bar!
add: Adds 2nd Bar Requests Console in Bar Storage (Next to the double-barreled shotgun)
qol: Moves the Newscaster, Request Console, Fire Alarm, and Intercom to new positions on the Bar's walls
/:cl:

### DIFF
--- a/_maps/map_files/PubbyStation/PubbyStation.dmm
+++ b/_maps/map_files/PubbyStation/PubbyStation.dmm
@@ -12411,10 +12411,7 @@
 	name = "Bar Lockdown";
 	req_access = list("bar")
 	},
-/obj/machinery/firealarm/directional/north{
-	dir = 4;
-	pixel_x = -4
-	},
+/obj/machinery/firealarm/directional/north,
 /turf/open/floor/iron/dark,
 /area/station/service/bar)
 "aVk" = (
--- a/_maps/map_files/PubbyStation/PubbyStation.dmm
+++ b/_maps/map_files/PubbyStation/PubbyStation.dmm
@@ -2022,7 +2022,7 @@
 /turf/open/floor/plastic,
 /area/station/security/lockers)
 "ahP" = (
-/obj/item/radio/intercom/directional/north,
+/obj/machinery/newscaster/directional/north,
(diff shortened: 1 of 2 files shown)


ackbarscrowbars:
bar
storage

SgtHunk (changes requested):
use the appropiate subtype for the firealarm, as that one has it's direction varedited

ackbarscrowbars:
> use the appropiate subtype for the firealarm, as that one has it's direction varedited

The fire alarm is like that with pubby to begin with, I only put it in a different place. 
I don't understand the problem. Does it cause any issue?

SgtHunk:
> The fire alarm is like that with pubby to begin with, I only put it in a different place. 
> I don't understand the problem. Does it cause any issue? 

It wasn't like that in the first place, it was an edit you made (as shown in files changed).
It causes issues, both in mapping standards and when futureproofing certain projects like wallening
In an ideal world, there shouldn't be varedits in maps and the TG codebase slowly tries to reach that goal

SgtHunk (approved):

PR #987, opened 2023-05-10
Allows beefcyto to actually spawn

Pepsilawn:
## About The Pull Request

Bro its been like half a year since carps arent simple animals whats up with that

## Why It's Good For The Game

I replaced the carp in the cell with a slime because carps are bitch to deal with nowadays and will just teleport outside the moated ruin + their cells can be gotten from moisture traps anyway
I also added a guaranteed fern plant because everything fun is locked behind them which says a lot because cyto isnt actually fun


Pepsilawn:
dude..

PR #989, opened 2023-05-14
Transfer honeycomb from bags into the all-in-one grinder

Peliex:
## About The Pull Request

Added an additional function call to the logic for dumping items from bags into the all-in-one grinder, this call checks to see if the bag contains any honeycombs and tries to fill the container with them.

## Why It's Good For The Game

This is a very simple quality of life change, being able to fill plant bags with honeycomb is already an interesting choice since they are not a plant, and having to take them out one by one to put into the reagent grinder is an unnecessary step.

## Changelog

:cl:
qol: dump honeycomb from bags into the all-in-one grinder
/:cl:


Peliex:
Closing this PR, moving changes to upstream

PR #1046, opened 2023-09-06
add pixel offsets so the collision makes sense

### LINKED ISSUE #893: Cherry trees are a single tile large object, with shadows baked in
Cherry trees don't behave correctly:
image
Only the origin point (red square) has collision when its one of the few places you should be able to walk into.
See-through component only works on the tile directly above it, none of the others that are still behind it proc the fadeout, since the object is technically only on that one square.
The shadow is baked into the object instead of being a plane thing, which makes it catch clicks and makes it hard to pick up any object under the tree's shadow on top of overlaying the shadow on top of mobs.

remove32:
## About The Pull Request
Pixel offset cherry tree. Any maps that use it might want to check that the positioning still makes sense!

## Why It's Good For The Game
Fixes #893
dreamseeker_HYoXcx5LuQ

## Changelog

:cl:
image: cherry tree now has appropriate collision
/:cl:

### DIFF
--- a/fulp_modules/mapping/station_objects/trees.dm
+++ b/fulp_modules/mapping/station_objects/trees.dm
@@ -3,3 +3,5 @@
 	desc = "A cherry tree that has been bio-engineered to keep its pink flowers year-round."
 	icon = 'fulp_modules/mapping/icons/cherry.dmi'
 	icon_state = "cherry"
+	pixel_x = -68
+	pixel_y = -20


remove32:
I believe the shadow being baked in is not something I can fix here, because I don't know how. Other examples of trees also had shadows baked in, so I'm not able to find a good example easily

PR #1054, opened 2023-09-30
Changes Cryo Cell color depending on temperature

T-for-Tea:
## About The Pull Request
A sheer cosmetic change, if the Cryo Cell is working with gas that is >460K, it will reflect on the color by turning it a fiery orange.
## Why It's Good For The Game
Currently, regardless of the temperature the Cryo Cell looks the same, even if you put pyroxadone in the beaker and set the temperature high, the only way to tell this is to either look up the cryo UI, or the tiny indicator on the thermomachine.

This change will make it very prominent that the Cryo is running hot (and likely containing pyroxadone considering someone set it up to work in high temperature.). Making pyroxadone is easy, but rarely done - it requires slime jelly which means xenobiologist must help medbay to produce this chemical. With a vivid visual change like this, this might encourage some cross departmental cooperation as well.

As the PR stands, it does not actively check the temperature to update the icon, rather it is updated whenever the state is changed (on, off, doors open or closed). If required, it shouldnt be too difficult to implement this, however its value should be evaluated as this will use a lot more processing power for a simple cosmetic change.

The color will change when it is above 460K, and the reason i picked 460K is that this seems to be the temperature at which pyroxadone works the best according to the TG wiki.

I've tested the change, it does work, it shouldnt interfere with any of the code inside cryo.dm, but for good measure i tried adding/removing beakers while the icon is being updated and there seems to be no issues. I also have a video, ill see if i can add it here somewhere.

https://github.com/fulpstation/fulpstation/assets/124678540/71b6a6ae-ddaa-4bf7-8242-aa77baa95376
## Changelog
:cl:
add: Cryo Cell now updates its sprite depending on temperature, if it is above 460K it will use a fiery orange sprite instead of the regular one

/:cl:


T-for-Tea:
first time ever touching BYOND code, so be warned. Not sure if this is the cleanest way to code this change, but it is what i got right now 🤡

PR #1082, opened 2023-12-10
FIXES Heliostation power

SgtHunk:
## About The Pull Request

Adds more cable redundancies to avoid power being shut off from a single mouse lmao (tell me if i need to add more and where idk i just connected maint cables)
Also probably fixes tcomms power issues!!!

## Why It's Good For The Game

Argue for the merits of your changes and how they benefit the game, especially if they are controversial and/or far reaching. If you can't actually explain WHY what you are doing will improve the game, then it probably isn't good for the game in the first place.

fa068b196cc805df152ae6e9dbf6b7dd

## Changelog

:cl:
fix: fixed helio power issues probably 
/:cl


JohnFulpWillard:
Hey Heliostation command hallway has 2 APCs and it breaks stuff

Especially since the one on the right is not connected to the powernet and when it runs out of power the whole hallway has no power

image

SgtHunk:
what the hell how did i forget to delete it

PR #1091, opened 2024-01-05
Fixes the mentor panel I think

JohnFulpWillard:
## About The Pull Request

I added more messages so in the future I can know exactly what isn't passing, but this works on my machine ™️ so I hope it works in-game too for non-admin mentors.

## Why It's Good For The Game

Mentor panel's Reply and Follow buttons actually working is pretty awesome tbh...


LuxAeternia:
Dear God, is it finally time? Time to say goodbye to the shitty, barely functioning Mentor Panel? I'd say it will be missed, but that's giving it too much credit.

Thanks a ton John! :D

DexeeXI:
Looks like this pull might fix an issue we're having over on Monke. Thanks for the link John!

PR #1100, opened 2024-02-01
Nanites

JohnFulpWillard:
## About The Pull Request

Adds Nanites as it was previously, but with some changes:

- Taking disks out uses Right-click instead of Alt-click
- Many messages sent to your chat are now balloon alerts
- It's now all fully modular
- The machines now can be deconstructed (and don't go invisible)
- No longer generate research points or put people on the crew monitor for free
- Does not have a public nanite chamber, people have to go get implanted manually by a Scientist
- Access to nanites have been moved from Robotics to Research
- Added more feedback and cool sounds, messes with speed on some machines to match its sound
- Nanite chamber now only takes one scanning module since it used to take both into account for its tiered part, making it immediately considered 'tier 2'.
- Mood effects are now only -/+ 7
- You can't upload more than one of the same protocol type
- All UIs now use TSX
- The Nanite Program Hub now works like other machines that gets designs through research, as in it will state '[x] new designs' and update its UI.
- People with nanites now generate 1.5 points instead of 3.5 & requires a client,

All TG edits is just me porting my own PR that's currently up on TG, we'll get it regardless next TGU but I needed it for Nanite points so I'm porting it with this PR, so no tg edit notes needed.

Heliostation (replaced the science range)
image

Selenestation (Replaced part of the Telescience maint room)
image

Pubbystation (Replaced the Experimentor room)
image

## Why It's Good For The Game

image

Nanites is a great subdepartment because it isn't as overcomplicated as Circuits but still has a use that helps others (rather than Circuit's limited innate selfish self-benefiting use) and contributes to the openness and sandbox of SS13 with many different types of interactions that takes a backside of the story.
Trying to take it away as a machine that you enter in a tube then forget about, now if you want nanites you have to go out of your way and get it implanted by another player, which is way better than previous (and is a trend with new stuff, such as the dna infuser), for benefits that may or may not turn bad throughout a round. Even if it's rare, it's still something that may occur, and that is enough of a downside to warrant its upsides (especially compared to something like Virology where it physically can't go bad once you've already gotten a 'good' disease).
Also toned down a bit, nanites now fits in the context of experisci and won't horribly throw everything out of balance. It would be great if we can get nanite-related exeperiments but I don't want to take this too far for the introductory PR, we can make more changes in the future, I just wanted to get it in first.

## Changelog

:cl:
add: Nanites
/:cl:


maxspells:
John Willard Standard Nanites™

SgtHunk:
you need to fix pubby decals and perhaps move the light/apc so there aren't 2 wall objects in the same wall it looks a bit ugly
other than that mapping lgtm

SGSDHSD-Enterprise:
hello the tgui text input box is really laggy and slow also the nanite chamber machine board doesnt have an icon on the circuit imprinter

JohnFulpWillard:
the nanite chamber board is the same issue as the circuit chamber and it's due to the sprite being larger than most (since it has animations for rising). I can think of a way to fix it.
The tgui text input box is an issue with tgui in general and there's nothing I can do but replace it with another function that would take away some of the experience for when it gets fixed

JohnFulpWillard:
icon patch added
image

Horatio22:
Hell yeah can't wait to blow everyone up with these bad boys

PR #1113, opened 2024-02-15
 Posession now does revive the dead and turns them into temporary vassals

FernandoJ8:
## About The Pull Request

What it says on the tin.
There's nothing else player facing (other than a balloon alert), but code-wise `can_make_vassal()` will now succeed even if the target is in hardcrit or worse. That check is now specific to the persuassion rack which is (currently) the only other way for someone to become a vassal.
## Why It's Good For The Game

It's a bugfix.

## Changelog
:cl:
fix: Posession now revives those who are dead or in hard crit and turns them into temporary vassals
qol: Added a balloon alert that indicates when someone cannot be tortured in a persuasion rack due to being dead or in hard crit.
/:cl:


OneAsianTortoise:
Uh I think Willard stole Fern's account

Horatio22:
@JohnFulpWillard  approved it - thanks!

PR #1125, opened 2024-02-28
pubby atmos edit

Sadboysuss:
## About The Pull Request

proper n2 and o2 to pure outlets, air mix  pipes are better now, layer manifolds on gas supply outlets. also flipped the direction of the science ordnance experiment thing cause it was the wrong way
## Why It's Good For The Game

Pubby is an old map and it's now updated to modern atmosian standards, instead of taking 5 minutes of your time to fix the atmos pipes, you get a proper room to work with as an atmos tech.
:cl:
qol: better atmos for Pubby
fix: fixes the direction of the science pressure experiment machine
/:cl:


Cure221:
No way, pubby 2 is real

Horatio22:
[15:59:05] Runtime in code/modules/atmospherics/machinery/components/components_base.dm,171: list index out of bounds
  proc name: set pipenet (/obj/machinery/atmospherics/components/set_pipenet)
  src: the omni oxygen filter (/obj/machinery/atmospherics/components/trinary/filter/atmos/o2)
  src.loc: the floor (141,94,2) (/turf/open/floor/iron)
 
 
 Check tile 141, 94

GuillaumePrata:
While the QoL changes with the multilayer pipes are good, this does makes it a lot messier to understand the pipework itself.

Something, something, new player friendly suggestion here:

pubby atmos feedback

If we swap the O2/N2 mixer to the other pink square, we can have the green filter line going in a straight line and not going under the windows.

I assume you will change the black O2 pipe too, but I will comment here just so more people don't do it later.

Sadboysuss:
image
added Guillaume's ideas

Sadboysuss:
:)

PR #1145, opened 2024-03-06
Makes round time actually work somewhat

JohnFulpWillard:
## About The Pull Request

this is a bandaid but server time starts at 19 hours and actually increments now.

It looks like 514/515 changed how time and suchs works and would require a lot more work than I'm willing to put into it rn.


remove32:
would... would giving it a -19 offset "fix" the problem?

JohnFulpWillard:
did that now

RealityOverseer:
Sir that did not fix it
image
proc/station_time minuses the offset, which means you are actually adding 19 hours.
proc/time_to_twelve_hour also minuses the offset.
At least that's what I think is causing the time to appear like above.

remove32:
i hate DST i hate DST i hate DST i hate DST i hate DST

PR #1160, opened 2024-03-15
Changes many decals all over the station

Bilbo367:
## About The Pull Request

A lot of small changes to decals all around the station.

## Why It's Good For The Game

Fixes some places where decals were missed, and makes some areas look more uniform in style to others. Also changes some other areas to improve the look of it.

Screenshot 2024-03-15 153642

image

image

Screenshot 2024-03-15 153822

### DIFF
--- a/_maps/map_files/PubbyStation/PubbyStation.dmm
+++ b/_maps/map_files/PubbyStation/PubbyStation.dmm
@@ -47715,6 +47715,12 @@
 	},
 /turf/open/floor/plating,
 /area/station/maintenance/department/cargo)
+"sSt" = (
+/obj/effect/turf_decal/tile/purple{
+	dir = 4
+	},
+/turf/open/floor/iron,
+/area/station/hallway/primary/aft)
 "sSA" = (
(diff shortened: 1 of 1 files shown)


Sadboysuss:
I think the yellow hazard strip is a bit overused...

Bilbo367:
> I think the yellow hazard strip is a bit overused...

I think this is valid for some cases like i dont like the cyro yellow strip or the one in sci maints, and on the lockers above that.

Horatio22:
I looked at the changes, and it looks fine, but just to confirm: you only used /tg/ decals, right? John keeps pubby up to date on the /tg/ map repo, so only /tg/ code can be used on this map.

Bilbo367:
> I looked at the changes, and it looks fine, but just to confirm: you only used /tg/ decals, right? John keeps pubby up to date on the /tg/ map repo, so only /tg/ code can be used on this map.

Yeah it should all be tg. I just checked and nothings under fulp. Assuming the fulp decals include some fulp.

PR #1162, opened 2024-03-16
Deathmatch : Monkey Business

sylvia-from-fulp-station:
## About The Pull Request
Here is the second and penultimate Deathmatch map from me.
Photoshop_242oZ1NbhT
## Why It's Good For The Game

Its a Death Match map where you are all monkeys. There isn't one like this, and it also tries to show some unique monkey gimmicks. Of my ideas, I found this by far the funniest. Celebrates this fix https://github.com/fulpstation/fulpstation/pull/1161
Chee ook.
## Changelog

:cl:
add: Adds koth_jungle.dmm and DM loadout: Monkey. In the Deathmatch menu it appears as "Monkey Business"


Horatio22:
This is probably going to be a headache to maintain down the road and would be better PR'd upstream.

PR #1168, opened 2024-03-17
Removes old nanite relics from the game

Bilbo367:
## About The Pull Request

Removes fluff, and the appropriate fluff stuff on helio

## Why It's Good For The Game

We have nanites so it don't make sense to have a relic site.


Horatio22:
Excellent fix. There's going to be a merge conflict with the other helio pr but otherwise I think this is great

SgtHunk (changes requested):
Good, but in this case you have to replace the maint room now with some other fluff theme, since its original design was meant to be a nanite lab

Bilbo367:
How about this?
Screenshot 2024-03-17 100031

Bilbo367:
How about dis?
image

SgtHunk (approved):

PR #1169, opened 2024-03-17
removes helio's layer 1 cables in sm chamber

JohnFulpWillard:
## About The Pull Request

This is a good idea on paper but layered cables is pretty non functional right now. You can place cables down there as normal and it would overwrite the SMES and behaves VERY weirdly, so I think it's best to avoid this altogether. I moved things around so SMES cables don't have to go through walls but still, no more layer 1 cables in the SM chamber.

I also removed the floating sm delam counter in CE's office and fixed the floating poster in telecomms


Pepsilawn:
I WARNED THEM ALL AND THEY IGNORED ME

PR #1171, opened 2024-03-17
Fixes Prisoner accounts and some refactors

Bilbo367:
## About The Pull Request

Refactors some code. and deletes ACCOUNT_PRISON so prisoners can have money.

## Why It's Good For The Game

Lets prisoners have money.

### DIFF
--- a/_maps/map_files/Heliostation/Heliostation.dmm
+++ b/_maps/map_files/Heliostation/Heliostation.dmm
@@ -1313,7 +1313,7 @@
 	name = "Supermatter Engine Room"
 	},
 /obj/effect/mapping_helpers/airlock/access/all/engineering/general,
-/obj/structure/cable/layer1,
+/obj/structure/cable,
 /turf/open/floor/plating,
 /area/station/engineering/main)
 "agx" = (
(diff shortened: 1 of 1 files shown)


fulp_modules/_defines/fulp_defines.dm
```diff
@@ -51,6 +41,3 @@
 
 ///Define for the 'Rabbits' Faction.
 #define FACTION_RABBITS "rabbits"
-
-//Define for the Syndicate Engineer Ruin, used in 'syndicate_engineer.dm"
-#define ROLE_SYNDICATE_ENGINEER "Syndicate Engineer"
```
JohnFulpWillard, line 56: What’s the reason for removing the define?

Bilbo367: We could keep it, its only used in one place so it seems weird to define it for only one use.

remove32:
filthy prisoners

fulp_modules/_defines/fulp_defines.dm
JohnFulpWillard: I think you should keep it in

Bilbo367:
let me fix my master

PR #1194, opened 2024-04-02
Vassel Recuperation Spam fix

Bilbo367:
## About The Pull Request

fix: Vaseel recuperation will no longer spam you when your stunned.

This spam:
Screenshot 2024-04-02 152450

## Changelog

:cl:
fix: Vaseel recuperation will no longer spam you when your stunned.
/:cl:


JohnFulpWillard:
This is a very bad fix and I tried looking into it and found massive issues with it being a subtype of cooldown actions, maybe we should revert it and go back to being our own action subtype.

PR #1197, opened 2024-04-05
Healing nanites works better on Lavaland

JohnFulpWillard:
## About The Pull Request

Regular healing nanite program now heals 0.3 per tick on the station, previously was 0.5
Advanced healing nanites now heals 1.2 a tick on the station. Previously was 1.5
You get the old healing if you're in low-pressure (space/lavaland/etc)

Also fixed the nanite dot icon not going away when you lose nanites (or make them stealth).

Mechanical repair not affected by this.

## Why It's Good For The Game

Nanites are a good tool that you can give miners to help them survive on Lavaland easier, while also being a useful tool on the station. This hopefully allows Nanites to keep its use in both cases while being better for miners than everyone else, as to encourage their use more and discourage its use as simply a tool to evade medbay.

### DIFF
--- a/fulp_modules/Z_edits/hud_edits.dm
+++ b/fulp_modules/Z_edits/hud_edits.dm
@@ -13,8 +13,8 @@
 
 /datum/atom_hud/data/diagnostic/basic/New()
 	. = ..()
-	hud_icons += list(NANITE_HUD, DIAG_NANITE_FULL_HUD)
+	hud_icons += list(DIAG_NANITE_FULL_HUD)
 
 /datum/atom_hud/data/diagnostic/advanced/New()
 	. = ..()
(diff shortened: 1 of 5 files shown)


remove32:
i'm hesitant to give shaft miners anything but i'm also not against nerfing nanites on station a bit. i also think its worth noting that the lavaland equipment check is something of a misnomer in my opinion - simply that helper checks if the pressure on your turf is less than 50. this change effectively buffs nanites anywhere unpressurized including space

JohnFulpWillard:
Yeah it's how I meant it to be, if you're running into space just to get nanites to heal you a bit more then you're either prepare for space or not really healing by much.

Science as a department to me has always been best when it revolves around mutual benefit with Mining, through upgrading their machines, Bags of Holding, mining tech, and other features that are no longer with us anymore.

PR #1199, opened 2024-04-06
Adds Protogens

SgtHunk:
## About The Pull Request

Adds protogens as an april-fools only species
They speak synthetic and have robotic limbs (but take 25% more overall damage)
Their organs are entirely cybernetic and so an EMP/Ion will be very punishing
Code-wise its very low maintenance, the most important addition is one (1) TG edit because of TGUI, but it opens up any additional species that might want to use secondary features instead of primary features (all this changes is where it's located in the prefs menu believe it or not)

## Why It's Good For The Game

i no longer have to keep adding them every april fools
+ it'll still be available if any admin wants to turn someone into a protogen in-game or smth like that

imagen
pictured: protofriends

## Changelog

:cl:
add: Species that starts with P and ends with rotogen
/:cl:


fulp_modules/features/species/protogen/protogen.dm
```diff
@@ -0,0 +1,78 @@
...
+	mutanttongue = /obj/item/organ/internal/tongue/robot
+	mutantstomach = /obj/item/organ/internal/stomach/cybernetic/surplus
+	mutantappendix = null
+	mutantheart = /obj/item/organ/internal/heart/cybernetic/surplus
+	mutantliver = /obj/item/organ/internal/liver/cybernetic/surplus
+	mutantlungs = /obj/item/organ/internal/lungs/cybernetic/surplus
+	mutanteyes = /obj/item/organ/internal/eyes/robotic
+	mutantears = /obj/item/organ/internal/ears/cybernetic
+	species_language_holder = /datum/language_holder/synthetic
+	changesource_flags = MIRROR_BADMIN | WABBAJACK | MIRROR_PRIDE | MIRROR_MAGIC | RACE_SWAP | ERT_SPAWN | SLIME_EXTRACT
```
JohnFulpWillard, line 30: If this is supposed to be admin-only / april fools-only then you probably shouldn't let players transform into them no?

PR #1214, opened 2024-06-03
Ball gown addition/first pull

tvesion:
Added ballgown to greyscale, clothing, skirtdress, and appropriate vendors.


Momo8289 (changes requested):
You've got a bunch of files in the `.github` directory that should not be there.

Momo8289:
Please fill out the PR template properly. Should include exactly what was changed/added, why, and a proper changelog.

fulp_modules/features/clothing/ballgown/autodrobe.dm
```diff
@@ -0,0 +1,251 @@
+/obj/machinery/vending/autodrobe
```
Momo8289, line 1: Instead of completely redefining the autodrobe, you should use the `Initialize` proc to add onto the list of products, then call the original `Initialize` proc. 

Something like this (i wrote this on my phone and its probably incomplete, don't just copy/paste):
```dreammaker
/obj/machinery/vending/autodrobe/Initialize()
    product_categories[4] += [/obj/item/clothing/under/dress/ballgown = 2]
    ..() // <- Calls the original definition of the proc
```

fulp_modules/features/clothing/ballgown/clothesmate.dm
```diff
@@ -0,0 +1,226 @@
+/obj/machinery/vending/clothing
```
Momo8289, line 1: Same thing as the comment I left on the autodrobe.

fulp_modules/features/clothing/ballgown/skirt_dress.dm
```diff
@@ -0,0 +1,115 @@
...
+
+/obj/item/clothing/under/dress/sundress
+	name = "sundress"
+	desc = "Makes you want to frolic in a field of daisies."
+	icon_state = "sundress"
+	custom_price = PAYCHECK_CREW
+	greyscale_colors = "#FFE60F#9194A5#1F243C"
+	greyscale_config = /datum/greyscale_config/sundress
+	greyscale_config_worn = /datum/greyscale_config/sundress/worn
+	flags_1 = IS_PLAYER_COLORABLE_1
```
Momo8289, line 103: This is all duplicate and should be deleted

Momo8289 (changes requested):

PR #1218, opened 2024-06-07
Turns Pubby's monastery lanterns on at round start

QuiteLiterallyAnything:
## About The Pull Request
This pull request turns Pubby Station's monastery lanterns on at round start. A few more of them have also been added for good measure as seen here:
Well-lit_Monastery

## Why It's Good For The Game
It seems that Pubby Station's lanterns were actually supposed the be on at round start since their 'icon_state' value was set to 'lantern_on'. This didn't actually turn them on though, so for ages players have been forced to tediously turn them on one by one every round if they want Pubby's monastery to look good. Even with them on they don't provide adequate lighting for the monastery, so this PR added a few more.

## Changelog

:cl:
qol: Pubby Station's lanterns are now turned on at round start— there are also more of them.
/:cl:

### DIFF
--- a/_maps/map_files/PubbyStation/PubbyStation.dmm
+++ b/_maps/map_files/PubbyStation/PubbyStation.dmm
@@ -22390,7 +22390,8 @@
 /area/station/service/chapel/asteroid/monastery)
 "bQe" = (
 /obj/item/flashlight/lantern{
-	icon_state = "lantern-on"
+	icon_state = "lantern-on";
+	light_on = 1
 	},
 /turf/open/misc/asteroid,
 /area/station/service/chapel/asteroid/monastery)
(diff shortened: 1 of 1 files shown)


Pepsilawn:
Can you turn them into subtypes instead the var edits are a bit silly

QuiteLiterallyAnything:
> Can you turn them into subtypes instead the var edits are a bit silly

Done— I think. . . Is it okay that I had to add something to tgstation.dme? Is there any simple way to do it otherwise?

fulp_modules/mapping/station_objects/lanterns.dm
```diff
@@ -0,0 +1,4 @@
+// Unique lanterns
+
+/obj/item/flashlight/lantern/lit_lantern
+	light_on = 1
```
Pepsilawn, line 4:
```suggestion
	start_on = 1
```

Pepsilawn (changes requested):

fulp_modules/mapping/station_objects/lanterns.dm
QuiteLiterallyAnything: Ah, alright.

PR #1232, opened 2024-07-10
Adds Bloodsucker brazier/adjusts candelabrum brightness

QuiteLiterallyAnything:
## About The Pull Request
This PR adds a new Bloodsucker lighting structure, the brazier. It is identical to the candelabrum though it's a bit brighter and it looks like this:
image

<sup>(I'm not experienced with making sprites so it's very basic.)</sup>

Also this PR makes the candelabrum brighter, puts the brazier and candelabrum under a new subtype of Bloodsucker structure for lighting structures, and makes Bloodsucker lighting structures turn off when unanchored (preventing them from being used outside of a Bloodsucker's lair.)
## Why It's Good For The Game
There are very few Bloodsucker structures as is and this PR adds a new, relatively simple one. 
## Changelog
:cl:
add: added the Bloodsucker brazier structure and increased the brightness of the candelabrum
fix: fixed a bug which allowed candelabrums to be left active outside of a Bloodsucker's lair
/:cl:


QuiteLiterallyAnything:
Low quality GIF of the newly added particle effects:
Behold_particles

fulp_modules/features/antagonists/bloodsuckers/code/structures/bloodsucker_crypt.dm
```diff
@@ -410,65 +407,84 @@
...
+	. = ..()
 	STOP_PROCESSING(SSobj, src)
-	return ..()
 
-/obj/structure/bloodsucker/candelabrum/update_icon_state()
-	icon_state = "candelabrum[lit ? "_lit" : ""]"
-	return ..()
+/obj/structure/bloodsucker/lighting/update_icon_state()
+	. = ..()
+	icon_state = "[name][lit ? "_lit" : ""]"
```
JohnFulpWillard, line 438: This should use ``base_icon_state`` instead of name

fulp_modules/features/antagonists/bloodsuckers/code/structures/bloodsucker_crypt.dm
```diff
@@ -410,65 +407,84 @@
...
+/obj/structure/bloodsucker/lighting/attack_hand(mob/living/user, list/modifiers)
 	. = ..()
 	if(!.)
 		return
 	if(anchored && (IS_VASSAL(user) || IS_BLOODSUCKER(user)))
 		toggle()
 	return ..()
 
-/obj/structure/bloodsucker/candelabrum/proc/toggle(mob/user)
+/obj/structure/bloodsucker/lighting/DblClick()
```
JohnFulpWillard, line 460: We have LMB/RMB/AltLMB/AltRMB, I don't think we should put this on double click cause it's like incredibly old and never used outside of orbiting people as a ghost.

fulp_modules/features/antagonists/bloodsuckers/code/structures/bloodsucker_crypt.dm
```diff
@@ -410,65 +407,84 @@
...
+
+/obj/structure/bloodsucker/lighting/proc/toggle(mob/user)
 	lit = !lit
 	if(lit)
 		desc = initial(desc)
-		set_light(2, 3, "#66FFFF")
+		set_light(active_light_range, light_power, light_color)
+		playsound(loc, 'sound/items/match_strike.ogg', 25)
+		if(name == "brazier")
+			particles = new /particles/brazier
```
JohnFulpWillard, line 476: This should be on the brazier subtype instead, same for the check below this one

fulp_modules/features/antagonists/bloodsuckers/code/structures/bloodsucker_crypt.dm
```diff
@@ -479,6 +495,23 @@
...
+
+// Candelabrum - Drains the sanity of non-bloodsuckers/non-vassals near it, has the brightness of a lightbulb
+/obj/structure/bloodsucker/lighting/candelabrum
+	name = "candelabrum"
+	desc = "It burns slowly and doesn't radiate heat."
+	icon_state = "candelabrum"
+	active_light_range = 3
+
+// Brazier - An alternate version of the candelbraum with roughly the brightness of a bonfire
+/obj/structure/bloodsucker/lighting/brazier
```
JohnFulpWillard, line 508: I don't think this should have the same effect as the Candelabrum, I've got potential ideas for what it can do but for now I think it would be good if you separated the effects of each onto the subtypes, they share light but insanity could be candelabrum's thing and brazier can do another effect.

JohnFulpWillard (changes requested):

fulp_modules/features/antagonists/bloodsuckers/code/structures/bloodsucker_crypt.dm
QuiteLiterallyAnything: Done— I think I implemented it correctly.

fulp_modules/features/antagonists/bloodsuckers/code/structures/bloodsucker_crypt.dm
QuiteLiterallyAnything: Switched to CTRL-click because regular Click() applies to _all_ forms of clicking, I can't find the proc for right-click, and alt-click always brings up the tile content's panel and I can't figure out how to override that.

fulp_modules/features/antagonists/bloodsuckers/code/structures/bloodsucker_crypt.dm
QuiteLiterallyAnything: Changed.

fulp_modules/features/antagonists/bloodsuckers/code/structures/bloodsucker_crypt.dm
QuiteLiterallyAnything: The effect has been removed from the brazier. Though, out of curiosity, what new effect(s) for it do you have in mind?

JohnFulpWillard: I was thinking of letting people buff their clan ability with it, like Nosferatu can temporarily gain the ability to ventcrawl, Tremere can unlock an ability's ascension or even use it as the place to upgrade powers at, etc.

QuiteLiterallyAnything: That sounds good! Please just remember that Bloodsuckers are likely to make multiple braziers for style points/aesthetic reasons. Given that, I think it shouldn't do anything overly specific, and it would be nice if the effect could stack with the number of braziers built. Above all else, please ensure that it's _fun_— thank you.

PR #1234, opened 2024-07-19
Removes Felinid brain damage

GuillaumePrata:
## About The Pull Request
Just does that.
The kitty ears keep the brain damage, fashion at your own risk.

I tried to add a new cringenid species to keep the feature around and it was not working, as I am on a tight schedule, cringenids might come back in the future!
## Why It's Good For The Game
It is not
## Changelog
:cl: Guillaume Prata
balance: Felinids don't have innate brain damage anymore.
/:cl:

PR #1246, opened 2024-09-05
Adds "Cateors" and Cateor Events

QuiteLiterallyAnything:
### The Cateor
This PR adds in a new, cat-themed meteor type: the "cateor.[^1]" Instances of this variant do very little structural damage to the station since they pass through everything with the exception of mobs. The mobs they do hit, however, undergo "catification." The full list of catification effects (as determined by mob type) is as follows:

1. Humanoids lightly explode and are turned into felinids.
2. Silicons get a random, unique ion law that (generally) makes them act like a cat.
3. Simple and basic mobs are just outright turned into cats.
4. Simple mobs that are already cats just double in size.

<details><summary>Also cateors look like this (expand to view image:)</summary>

Cateor
(Nyan Cat was a partial inspiration/reference for this sprite.)

</details>

### Cateor Events
Cateors do not appear in regular meteor storms. Instead, they are limited to unique "cateor storms" and "stray cateor" events. The former can only be triggered by wizards and mischievous administrators, while the latter only spawns a singular cateor that targets a random crew member. Cateor storms also trigger a new joke alert level which lasts for their duration, causing extreme shuttle delays (ETA multiplier of 255, which is only offset by the fact that this alert level should only be in effect for ninety seconds at a time.)

### Cateor Spell
Wizards can apply cateors directly through the new spell "Direct Cat Meteor" (spell demonstration video posted in PR comments.) This spell also has a unique scope for long-range usage.
## Why It's Good For The Game

- More round event variety.
- It's (hopefully) funny.
- It has minimal _direct_ effect on overall game balance.
## Changelog
:cl:
add: Added cateors
add: Added "cateor storm" and "stray cateor" events, with the former being purchasable by wizards!
/:cl:
[^1]:(It was the first name that came to mind. I know "meowteor" sounds better, but I am not immune to sunk-cost fallacy.)


Sadboysuss:
This is really cool but you're pring this to a codebase that despises felinids so it's a bit silly.
Additionally I don't think it fits as a meteor, it would make sense for a different event imo, like a mob doing this.

QuiteLiterallyAnything:
> This is really cool but you're pring this to a codebase that despises felinids so it's a bit silly. 

I've just reviewed Fulp's history of felinid-related PRs. If I'll be honest, it all seems to be just a tad disingenuous. Aside from the fact that felinids are effectively being readded with #1234, all of the measures previously taken to remove them seem half-hearted and minimally effective (e.g., removing felinids but not removing cat ears.)

I understand that catpeople can be difficult to integrate into a minor-friendly environment, especially since they are more often than not prone to inappropriate interpretation. Still, I think there are ways to fix this problem without outright deleting them.

> Additionally I don't think it fits as a meteor, it would make sense for a different event imo, like a mob doing this.

"Catifying magicarp" could exist in tandem with this.

QuiteLiterallyAnything:
Video demonstration of the newly added "Direct Cat Meteor" spell:

https://github.com/user-attachments/assets/e3d3520a-536e-4e90-af9c-b2b2759a2210

Pepsilawn:
Sound effects need some kind of attribution somewhere

QuiteLiterallyAnything:
> Sound effects need some kind of attribution somewhere

Done

PR #1259, opened 2024-10-11
Adds a basic cateor smite

QuiteLiterallyAnything:
## About The Pull Request
This PR adds a new admin smite which spawns a cateor near a target and launches it at them (with additional target immobilization options.) This probably isn't coded very efficiently, but it is by all means functional. A video demonstration of the new smite can be found through the Youtube link [here](https://www.youtube.com/watch?v=Am4zrSS626g).
## Why It's Good For The Game
It's a fun thing for admins to interject into a round regardless of context.
## Changelog
:cl:
add: added a new admin smite which involves cateors

### DIFF
--- a/fulp_modules/features/admin_chicanery/smites/catification.dm
+++ b/fulp_modules/features/admin_chicanery/smites/catification.dm
@@ -5,11 +5,30 @@
 
 	/// Boolean indicating whether or not we stun the target before launching a cateor at them.
 	var/stun_target = TRUE
+	/// Boolean indicating whether or not the user has accepted the smite's disclaimer.
+	var/warning_accepted = FALSE
 
 /datum/smite/catify/configure(client/user)
 	. = ..()
+	/// Warn the admin about the exact function of the smite.
+	warning_accepted = input(user,
+						"This smite will spawn a cateor at a random point within a roughly three \
+						tile range of its target. The cateor doesn't have perfect homing, and it is \
+						liable to hit any mob between it and the target. Please ensure that no mobs \
+						are present near the target unless you wish to risk hitting them.",
+						"DISCLAIMER",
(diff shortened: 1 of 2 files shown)


fulp_modules/features/admin_chicanery/smites/catification.dm
```diff
@@ -0,0 +1,50 @@
...
+	/// Copied over with adjustment from 'pandora.dm', gets a spawnpoint for our cateor at a certain
+	/// distance from the target.
+	for(var/t in spiral_range_turfs(3, T))
+		if(get_dist(t, T) > 1)
+			valid_turfs.Add(t)
+
+	/// Spawn the cateor and launch it.
+	var/obj/effect/meteor/cateor/new_cateor = new /obj/effect/meteor/cateor(pick(valid_turfs), get_turf(target))
+	addtimer(CALLBACK(new_cateor, PROC_REF(Destroy)), 15 SECONDS)
+	new_cateor.chase_target(get_turf(target), home = TRUE)
```
remove32, line 50: Tested this - it does work but it will impact any other valid target that happens to be in the way

QuiteLiterallyAnything: I don't know if there's any particularly easy way to prevent that, so I've added a disclaimer to the smite and shortened its automatic self-destruction timer down to five seconds.

remove32: realistically, fixing that would involve making a different version of cateor which only collides with the target. It's less of something I want you to change and more something I think was worth pointing out

QuiteLiterallyAnything: Okay

remove32:
we can always remove the warning or streamline it later on. looks good

PR #1272, opened 2024-11-13
fixes issue from opendream testing

JohnFulpWillard:
## About The Pull Request

Was trying to run fulpstation in opendream and thought these might be worth messing with.


remove32:
im looking forward to doing more testing with opendream in a bit. busy moving atm.

PR #1290, opened 2024-12-20
Splits off our defines and maps from fulp_modules

FernandoJ8:
## About The Pull Request

Moves the files previously under `fulp_modules/_defines` to `code/__DEFINES/fulp_defines`
Moves the files previously under `fulp_modules/_maps` to `_maps/fulp_maps`
Updates our contributing [README](https://github.com/fulpstation/fulpstation/tree/master/fulp_modules#readme) accordingly.
## Why It's Good For The ~~Game~~ Code

Defines: This allows us to make use of our defines in TG edits and other fulp code that might, for one reason or another, need to exist without the `fulp_modules` folder.

Maps: Allows UpdatePaths to detect our files without changes to the script. The damage is minimal regardless, since they get their own folder and we are already forced to keep our stations outside of `fulp_modules`
**edit:** ...and allows linters to check maps they weren't before. Good.

No changelong, since there's nothing player-facing.


JohnFulpWillard:
can you fix the mapping errors too I don't wanna merge this until it's done cause every other PR will start to fail

JohnFulpWillard:
Nvm saw the PR now. You'll have to fix conflicts.

PR #1303, opened 2024-12-26
[tm only] TGU staging branch

Sadboysuss:
## About The Pull Request
I suspect the original pr can't be merged because there is over 10000 commits, this is all of them squash merged into one, let's see if this works.


fulp_modules/features/antagonists/bloodsuckers/code/powers/feed.dm
```diff
@@ -49,6 +49,8 @@
 	return TRUE
 
 /datum/action/cooldown/bloodsucker/feed/DeactivatePower()
+	if(!active)
+		return
```
JohnFulpWillard, line 53: This should never be the case, this is hiding an issue. Remove please.

QuiteLiterallyAnything:
A couple things I noticed while playing on the server when this was test-merged:

1. Heliostation's public mining shuttle seems to have been replaced with regular one (or is otherwise not loading properly).
2. Round end sound effects/music now seem to include Fulpstation's round start music.

FernandoJ8:
> A couple things I noticed while playing on the server when this was test-merged:
> 
>     1. Heliostation's public mining shuttle seems to have been replaced with a regular one (or it's otherwise not loading properly).
> 
>     2. Round end sound effects/music now seem to include Fulpstation's round start music.

On 1: It was previously the mining shuttle used by Northstar, which has been removed along with the map itself. Does it have much importance? We could always re-add it as helio's own mining shuttle, but if there isn't a good reason for it, I'd rather use the standard variant.

FernandoJ8:
> A couple things I noticed while playing on the server when this was test-merged:
> 
>     1. Heliostation's public mining shuttle seems to have been replaced with a regular one (or it's otherwise not loading properly).
> 
>     2. Round end sound effects/music now seem to include Fulpstation's round start music.

On 2: That's config as far as I can tell, so admin issue.

fulp_modules/features/antagonists/bloodsuckers/code/powers/feed.dm
Sadboysuss: done

PR #1306, opened 2024-12-29
Gives fulp its own changelog files (...again)

FernandoJ8:
## About The Pull Request

Makes it so that the changelogs for our PRs are saved to the ``fulp_modules/data/html/changelogs`` folder, to be preserved across TGUs separately from TG's,

Additionally edits TG's ``compile_changelogs`` workflow to apply to ours. To enable it, a repository secret called "ACTION_ENABLER_FULP" must be created. (See ``.github\workflows\compile_changelogs.yml``, ln 16)
## Why It's Good For The Game

It's good to have our own changelogs. Paves the ground for QLA's project to add our own changelog screen as well.
## Changelog
:cl:
server: Gives fulp its own changelog folder for preservation.
/:cl:


Sadboysuss:
With the new tgu I did, changelogs will be preserved anyways if we do tgu properly from now on

FernandoJ8:
> With the new tgu I did, changelogs will be preserved anyways if we do tgu properly from now on

That is true, but we would get collisions if we didn't separate them. The file name for each changelog is just "AutoChangeLog-pr-[pr number].yml".
Plus, if we want to implement a separate changelog menu, which we likely do, it's better to separate them anyway.

.github/workflows/compile_changelogs_fulpstation.yml
```diff
@@ -0,0 +1,73 @@
+name: Compile changelogs (FULP)
```
JohnFulpWillard, line 1: What does this do that makes it worth doing this over a tg edit?

FernandoJ8: Nothing in particular, besides avoiding the edit. Would you rather I just edit compile_changelogs?

FernandoJ8: Replaced it with a TG edit.

remove32:
im not 100% on how this works (workflows still confuse me a bit) but im willing to give this a go

PR #1333, opened 2025-01-21
Fixes/Simplifies the "Cateorize" Admin Smite

QuiteLiterallyAnything:
## About The Pull Request
This PR fixes the (currently nonfunctional) "cateorize" smite by simplifying it. Previously the smite worked by spawning a cateor near a target and directing it towards them, but with this PR it'll function by spawning a cateor _at_ a target to instantaneously hit them.

## Why It's Good For The Game
The cateorize smite isn't working for some reason, and I understand neither why that is nor what has caused it. It seems to be something involving an error in what I could only vaguely (and perhaps nonsensically) describe as "movement loop override logic". My brief attempts at debugging the issue have failed, and this solution will fix it while also making the smite much more reliable overall.

## Changelog

:cl:
admin: Fixed the cateorize smite by simplifying it.
/:cl:

### DIFF
--- a/fulp_modules/features/admin_chicanery/smites/catification.dm
+++ b/fulp_modules/features/admin_chicanery/smites/catification.dm
@@ -1,72 +1,9 @@
 /// Catification smite
-/// Allows an admin to spawn a cateor near a target and direct it at them.
+/// Allows an admin to spawn a cat meteor at a target's location and instantly hit them with it.
 /datum/smite/catify
 	name = "Cateorize"
 
-	/// Boolean indicating whether or not we stun the target before launching a cateor at them.
-	var/stun_target = TRUE
-	/// Boolean indicating whether or not the user has accepted the smite's disclaimer.
-	var/warning_accepted = FALSE
-
-/datum/smite/catify/configure(client/user)
-	. = ..()
-	/// Warn the admin about the exact function of the smite.
-	warning_accepted = input(user,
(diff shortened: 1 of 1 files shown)


JohnFulpWillard (changes requested):
Doesn't this just have the exact same effect as the purrification smite?

QuiteLiterallyAnything:
> Doesn't this just have the exact same effect as the purrification smite?

Possibly, but I'm not seeing that as an option in the smiting menu on a debug build of the server.

JohnFulpWillard:
It's in the VV panel instead for some reason, perhaps a good PR to make upstream?

image

QuiteLiterallyAnything:
> I feel these are easy enough to replicate ingame with admin tools that it doesn't require a dedicated button

True, it only takes spawning a cateor and assuming control of it to replicate. Still, it's very precise (as in it can target a specific individual in a crowd) and could just be convenient for any admin wanting to hit a lot of people with cateors. @maxspells originally requested it with [this](https://discordapp.com/channels/279359975250198528/617105911168892928/1293967062691086428) Discord message, and @GuillaumePrata might've been interested in it based off of preceding context. I'll ping them both with a link to this PR to get their opinion.

Do you have some other underlying concern (worries about continued maintenance of the feature, this PR being a hassle, or something similar)? I don't understand how keeping the option would be a detriment.

PR #1334, opened 2025-01-21
Makes Bloodsucker Abilities Part of a Tab on the Status Panel

QuiteLiterallyAnything:
## About The Pull Request
This PR gives bloodsucker abilities their own action subpanel/tab as part of the typical SS13/BYOND status panel (the one that's always on the side of the screen and has a lot of clickable text commands). This functions in much the same way as it does with spells (the tab is only given to people who receive the particular abilities associated with it, et cetera).
## Why It's Good For The Game
These changes allow bloodsuckers to use their abilities even if their ability HUD buttons cease to function for whatever reason. 
## Changelog
:cl:
qol: Bloodsucker actions are now part of a subpanel/tab within the standard BYOND game status panel.
/:cl:

### DIFF
--- a/fulp_modules/features/antagonists/bloodsuckers/code/powers/_powers.dm
+++ b/fulp_modules/features/antagonists/bloodsuckers/code/powers/_powers.dm
@@ -7,6 +7,7 @@
 	button_icon_state = "power_feed"
 	buttontooltipstyle = "cult"
 	transparent_when_unavailable = TRUE
+	panel = "Bloodsucker"
 
 	/// Cooldown you'll have to wait between each use, decreases depending on level.
 	cooldown_time = 2 SECONDS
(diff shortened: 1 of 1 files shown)


JohnFulpWillard:
We have been trying to move away from the stat panel, this probably isn't a great idea

QuiteLiterallyAnything:
> We have been trying to move away from the stat panel, this probably isn't a great idea

That's entirely fair. Now that I think about it, this entire PR has just been an indirect response to an issue I've noticed with the HUD buttons system— primarily _**THIS THING:**_

image

I can't recreate it on a debug build of the codebase, but on multiple occasions in the past (during live games) this button manager bar has managed to irreversibly destroy my client's HUD buttons. One will somehow disappear when dragged near it, and then it destroys **all** of them when alt-clicked to "reset" everything. In recent memory this has only happened when I've been playing bloodsucker, so I don't know if it's something involving this antagonist specifically or a larger /tg/ issue.

This PR might be a good temporary solution to this (poorly documented, possibly isolated, and generally enigmatic) issue, but I also agree with the prospect of closing it. My only qualm is that I don't have enough information to resolve the issue otherwise by opening up a report with /tg/.

remove32:
i think as long as there is support for them in the games UI, having them available in the stat panel is a decent backup. yes stat panel sucks for a variety of reasons but its not going away anytime soon without significant redesign of UI elements, so i think we can go ahead with this as long as we acknowledge UI as the primary approach

PR #1348, opened 2025-02-02
Adds fulp space law

JohnFulpWillard:
## About The Pull Request

Looks like this

image

~~Would like to remove the banner and stuff first before it gets added.~~
turns out there is no way to do so.


JohnFulpWillard:
Yeah cause we should link the TG wiki for most everything else

PR #763, opened 2022-08-04
Ports the Bruhspace Anomaly

JohnFulpWillard:
## About The Pull Request

Ports https://github.com/TheSwain/Fulpstation/pull/81
Makes it a little better, it no longer affects non-sentient mobs, so it isn't spamming logs.

## Why It's Good For The Game

bruh

PR #1363, opened 2025-02-18
Removes the Bruhspace Anomaly Event

QuiteLiterallyAnything:
## About The Pull Request
This PR removes the bruhspace anomaly event (originally made in [this PR](https://github.com/TheSwain/Fulpstation/pull/81) and ported to our current repository through #763).
## Why It's Good For The Game
The bruhspace anomaly event is good as a concept, but it doesn't seem fully developed. Currently it just makes every living player say "bruh" at a random point in time. This can become repetitive to players after a while, and since there's nothing to actually prompt the event, it rarely coincides with anything meaningful. 

If the event triggered _because_ of something worthy of making every living player say "bruh" then that would be perfect. As it stands, though, the event just uses up an event roll opportunity to no particular end.
## Changelog
:cl:
del: Removed the bruhspace anomaly event from standard rotation and made it only triggerable by administrators.
/:cl:


JohnFulpWillard:
We could probably just remove this, I re-added it in the new repo because I find it funny to have it play ironically, as a way of showing NT's jokes as outdated/out of time, and also as a way of taking up some event space to make meteors happen a little less frequently because (at the time, i dont really know if its still the case) it was really one of the few things that caused a round end.

QuiteLiterallyAnything:
> We could probably just remove this, I re-added it in the new repo because I find it funny to have it play ironically, as a way of showing NT's jokes as outdated/out of time, and also as a way of taking up some event space to make meteors happen a little less frequently because (at the time, i dont really know if its still the case) it was really one of the few things that caused a round end.

I do think it has a lot of potential as an admin "secret", but I can make this a full removal if you'd like. The problem with meteors could be resolved by overriding the event's 'weight' var if it ever becomes an issue again.

JohnFulpWillard:
Yeah just remove it

QuiteLiterallyAnything:
> Yeah just remove it

Alright, since/if you're **_absolutely sure_**— done.

PR #398, opened 2021-10-30
Makes the AI less horny

Horatio22:
## About The Pull Request

Removes the lines from txt files that go against our server policies on hate speech and sexual content. I don't think this should cause any conflicts with TGUs, but who knows (not me).
 
## Why It's Good For The Game

It's weird banning people for ERP while the AI has a law that makes it talk about fetishes. Sends mixed signals.

## Changelog

:cl:
del: Removed text lines that break server rules.
/:cl:

PR #1377, opened 2025-03-08
Removes/Overrides Cards Against Spess

QuiteLiterallyAnything:
## About The Pull Request
This PR completely removes Cards Against Spess as both a card deck item and a purchase option in entertainment vendors.
## Why It's Good For The Game
Cards Against Spess (being a reference to [Cards Against Humanity](https://en.wikipedia.org/wiki/Cards_Against_Humanity)) might be too inappropriate for a minor friendly server such as Fulpstation. It would be possible to just remove the more inappropriate card options from 'strings/cas_black' and 'strings/cas_white' (as attempted in PR #398) but this approach has a few advantages:

1. It's modular.
2. We don't need to worry about the possibility of our upstream editing the string files in the future.
3. We don't need to ponder the ethical implications of keeping a secondhand explicit reference in our codebase.
## Changelog
:cl:
del: Removed Cards Against Spess as both a card deck item and a purchase option at entertainment vendors.
/:cl:


JohnFulpWillard:
Don't think I want this, CAS is a great card game and is probably the one I've played the most in-game, especially during now lowpop rounds getting everyone for a game of CAS sounds like a fun idea

QuiteLiterallyAnything:
> Don't think I want this, CAS is a great card game and is probably the one I've played the most in-game, especially during now lowpop rounds getting everyone for a game of CAS sounds like a fun idea

That's fair. Still, the current version has a _lot_ of explicit content. I suppose it might be possible to override it by creating a new pair of 'fulp_CAS' files and replacing all of the explicit stuff with generic equivalents (in the style of Pokemon's "[jelly-filled donuts](https://www.youtube.com/watch?v=gAbXcN5PvhA)"). Thoughts? That might actually make it funnier...

PR #1380, opened 2025-03-09
Implements More Coffin Dirs

QuiteLiterallyAnything:
## About The Pull Request
This PR adds more coffin dirs by flipping existing coffin sprites and implementing them appropriately
## Why It's Good For The Game
This allows bloodsuckers to have lairs with coffins facing both left and right!
## Changelog
:cl:
qol: Made coffins able to face both left and right.
/:cl:


QuiteLiterallyAnything:
Drafting for a bit because I think something messed up with the sprites...

QuiteLiterallyAnything:
Nevermind, it seems fine.

JohnFulpWillard:
did you forget to do the same for all the other coffin types?

QuiteLiterallyAnything:
> did you forget to do the same for all the other coffin types?

No, it just doesn't seem to be showing up on IconDiffBot check for some reason. It seems like it shows up on the actual file though (even when downloaded from GitHub).

JohnFulpWillard (changes requested):
Actually can you try to add these to TG first?

QuiteLiterallyAnything:
> Actually can you try to add these to TG first?

I might be able to with the regular coffin, but TG doesn't have the others (I think). I also don't have any TG GBP, so if adding the non-standard coffins would count as a "feature" then it might not work out.

JohnFulpWillard:
Ported upstream

PR #1264, opened 2024-10-25
october tgu (sylvia blue name app)

sylvia-from-fulp-station:
october tgu

its good for the game because uhh

-5000 sound files moved
-new antag 
-pirate RP
-cats go meow
-being evil
-re added buckshot

alot more stuff too

thanks to john willard, anette, grungus and fern for their help

PR #1386, opened 2025-03-21
[MDB IGNORE] [IDB IGNORE] March '25 TGU

JohnFulpWillard:
## About The Pull Request

Beefmen clothing is still broken but there's an open PR on TG to fix that, so hopefully it'll be done by next TGU
Brings the exiled king code/sprites back, commented out for now as it's broken. It was removed in Sylvia's TGU https://github.com/fulpstation/fulpstation/pull/1264 which if it was broken it still should've been commented out. Maybe this is gonna get put upstream in some form so maybe we won't need it afterall but at least we shouldn't forget about it.

MDB and IDB ignore because it's been quite a while since last TGU and I don't want to break diffs

## Changelog

:cl:
server: March '25 TGU
/:cl:


sylvia-from-fulp-station:
Kobsa asked for the king to be removed

QuiteLiterallyAnything:
> Kobsa asked for the king to be removed

That's entirely fair, but where?

PR #1410, opened 2025-04-21
Misc fixes and additions to beef_cytology.dmm

DoctorNut:
## About The Pull Request

Adds some missing stuff and small balancing to beef_cytology.dmm. Namely the missing growing vat and some mining stuff. Also makes the vendors free.

## Why It's Good For The Game

Currently the ruin cannot be used for its intended purpose, as it lacks a growing vat. It also has no way to refine plasma, as ghetto refinement methods generally destroy plasma ore. The vendors here need to be free as the ghost role has no ID and cannot make money. This adds an airlock leading to a section of the ruin with a growing vat. Some mining gear to make collecting ore easier as well as a redemption machine to allow plasma to be properly refined are also now present. Added a fire extinguisher to the specimen chamber in case the slimes are freed and a player doesn't already have one.

## Changelog

:cl:
balance: Added miscellaneous mining equipment to beef_cytology.dmm to make mining a little easier and plasma refinement possible
balance: Added an extra fire extinguisher to the specimen chamber in beef_cytology.dmm, just in case
map: Added an airlock to an area of beef_cytology.dmm that now has a growing vat, as well as moved some lockers and pipes around
fix: beef_cytology.dmm vendors are now free and the jukebox no longer requires an ID with bar access.
/:cl:

### DIFF
--- a/_maps/fulp_maps/RandomRuins/IceRuins/beef_cytology.dmm
+++ b/_maps/fulp_maps/RandomRuins/IceRuins/beef_cytology.dmm
@@ -9,6 +9,10 @@
 "ak" = (
 /turf/open/misc/asteroid/snow/standard_air,
 /area/icemoon/underground/explored)
+"ap" = (
+/obj/structure/extinguisher_cabinet/directional/east,
+/turf/open/floor/mineral/plastitanium/red,
+/area/ruin/powered/beefcyto)
 "av" = (
 /turf/open/floor/eighties,
 /area/ruin/powered/beefcyto)
@@ -29,6 +33,14 @@
 /obj/structure/flora/rock/pile/icy,
 /turf/open/misc/asteroid/snow/icemoon,
 /area/icemoon/underground/explored)
+"bd" = (
+/obj/machinery/door/airlock/external/glass/ruin,
+/obj/effect/mapping_helpers/airlock/cyclelink_helper{
+	dir = 8
+	},
+/obj/structure/fans/tiny,
+/turf/open/floor/plating,
+/area/ruin/powered/beefcyto)
 "bg" = (
 /obj/machinery/light{
 	dir = 8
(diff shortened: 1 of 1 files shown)


DoctorNut:
placement of the ORM is ugly, but I couldn't think of anywhere to put it that wouldn't make me change the map more

QuiteLiterallyAnything (commented):
A few points:

1. You could try placing the ORM in the ruin's maints surrounded by a couple of burnt decals and a note var-edited to read "Для плазмы пла́вка" ("for plasma melting" roughly translated via [Russian dictionary](https://en.openrussian.org/)). Also, are you sure that the ORM won't automatically link to the station's silo at round start?
2. These pipes will be a bit less accessible if they go under the kitchen machinery: <br>image
3. Consider moving this moisture trap over to this spot in order to remedy the abnormal bit of negative space by the door:
suggestion

DoctorNut:
So I can't seem to figure out how to make the ORM not connect to the silo on roundstart. I'm going to just replace it with a furnace and console like oldstation.dmm has. Everything is now working though.

QuiteLiterallyAnything (commented):
This looks good, but there is one minor thing that I forgot to mention in my last review. If the growing vat is left outside then every mob made with it will be exposed to the cold. This will be bad for slimes especially. 

<details><summary>You could try moving it over to this area.</summary>

image

</details>

The airlock should work regardless of if there's a growing vat outside it. If you wanted to make it look a bit less aesthetically abrupt then you could put caution line decals under its doors and maybe add a lightbulb.

DoctorNut:
Should I include containment for the cytology vat and potentially cameras and a slime management console?

DoctorNut:
Does this work for the growing vat? 
image

QuiteLiterallyAnything:
> Should I include containment for the cytology vat and potentially cameras and a slime management console?

This would give the ruin more of a xenobiology focus, and the entire spawner would then require a lot of other adjustments. You _could_ do it, but something like that might be large enough for its own PR (separate from the partially unrelated fixes you intend to make here).

> Does this work for the growing vat?

I'm not sure— it might contrast too much with the rest of the room (though that could be a good thing since it would make the vat a focal point). Regardless, consider removing the purple tile decals that are overlapping weirdly with the reinforced flooring.

DoctorNut:
Made requested changes and added a note roughly translating to "Slime does not like water" in the shower in the vat room.

PR #1427, opened 2025-10-31
[MDB IGNORE] [IDB IGNORE] WIP TGU

JohnFulpWillard:
## About The Pull Request

This is a big ass PR due to Fulp being dead for several months prior to this. Server reopening was kinda in the blue so we're scrambling to get this together.


FernandoJ8:
ok so it seems we're going to have to do this for a few files
0 line changes will save the fulpstation repo

QuiteLiterallyAnything:
This seems to have broken the Fulpstation changelog implemented in #1345 (and hotfixed in #1372).

PR #1488, opened 2026-05-30
Different test for fulp module schemas

JohnFulpWillard:
## About The Pull Request

testing this as an alt to https://github.com/fulpstation/fulpstation/pull/1308 that'll have 1 less tg edit.


JohnFulpWillard:
Pushed usptream instead

PR #1497, opened 2026-08-04
Mentors 2.0

JohnFulpWillard:
## About The Pull Request

Mentorhelp now uses TGUI input
Mentors now have their own spot in the player list
Unfollowing someone as a Mentor is done through Escape rather than a verb
MentorSay (F4) is now a TGUI say box
Mentorhelp is now in the Escape menu
Mentors using Mentorhelp has a submenu to allow them to access Mentor Manager
Removes Head of Pseudostaff
Moves Mentor verbs to Admin panel under a Mentor submenu
Removes double "Current Mentors" from Mentorwho

This is what Mentors can see

image

Admins

image

Players (Without the Readmin button)

image

image

## Why It's Good For The Game

This is quite a bit of TG edits but I think it has become necessary to move our stuff to TGUI, players and mentors can't suffer continuously being held back from TGUI especially now that the Stat panel is being removed.

I hope eventually we can push this upstream and not have to deal with conflicts, like how modchat already exists there, but I don't think it'll happen cause it's a downstream issue through and through.

## Changelog

:cl:
add: A lot of Mentor stuff now uses TGUI and has been moved away from the Stat panel.
/:cl:


fulp_modules/features/mentors/mentor.dm
```diff
@@ -34,6 +32,27 @@ GLOBAL_PROTECT(mentor_href_token)
...
 		GLOB.mentors += owner
 
+/datum/mentors/proc/mentor_follow(mob/living/followed_guy)
+	if(isnull(followed_guy))
+		return
+	owner.mob.reset_perspective(followed_guy)
+	to_chat(GLOB.admins, span_adminooc("<span class='prefix'>MENTOR:</span> <EM>[key_name(owner.mob)]</EM> is now following <EM>[key_name(followed_guy)]</span>"))
+	to_chat(owner.mob, span_info("Use \"Escape\" to stop following [key_name(followed_guy)]."))
+	log_mentor("[key_name(owner.mob)] began following [key_name(followed_guy)]")
+	RegisterSignal(owner.mob, COMSIG_MOB_KEYDOWN, PROC_REF(mentor_unfollow))
```
JohnFulpWillard, line 45: Should register for when perspective is reset in case someone uses the OOC verb.

JohnFulpWillard:
Thinking about TGUI-Say being turned off, I don't think Asay works either so maybe worth pushing upstream after i look into it.

PR #1508, opened 2026-08-18
Adds NTSL & The Network Admin

JohnFulpWillard:
## About The Pull Request

Ports https://github.com/Monkestation/Monkestation2.0/pull/2199 which itself is a port from https://github.com/yogstation13/Yogstation/pull/4217

You can now upload scripts to Telecommunications which affects how people speak

image

And you got a cool guy who came along with the computer. Apparently the operator comes with the computer, some deal with the company that makes the software they say.

image

(hud item not included and only shown for secHUD demonstration purposes)

Also hello Fulp exclusive jobs, we lost you there for a while. Welcome back.

This doesn't come with Yog AI rework which is partially what Network Admin's job revolves around, but it's fine enough as-is that I don't think we necessarily need it.

Config files are being included in this so people porting this can get it, but because we're not maintaining it as part of the Fulpstation downloadable, it'll be removed next TGU and instead become server-side only.

## Why It's Good For The Game

New engineering job and content from the now-defunct Yogstation at a moderate price? I suppose so.

## Changelog

:cl: Altoids1, Bibby0110, TheReddDragon, Hopekz, ReddicusDragon, Lostlarry, Amelia0010, monster860, iamthesilly, Gbobster-0, adamsong, BurgerLUA, JohnFulpWillard
add: Adds NTSL to Telecommunications, and the Network Admin who's job is to manage it.
/:cl:


remove32:
I'm somewhat concerned about the depth of play that the network admin could possibly achieve. their entire purpose is to... what, fuck around with comms? Clutter it by putting Poly onto general comms? The whole thing seems like it might be better suited as a function of chief engineer to be honest.

JohnFulpWillard:
CE exists to oversee Engineering so I don't think giving them 'custom' job content would really benefit what we want out of the role.

PR #1509, opened 2026-08-18
Is this aura (WIDER screen)

JohnFulpWillard:
## About The Pull Request

Increases our widescreen size from 19x15 to 23x15

## Why It's Good For The Game

Let's try it bros

## Changelog

:cl:
balance: There's now 4 more horizontal tiles.
/:cl:


JohnFulpWillard:
Testing shows this is not a great idea.

remove32:
aurant

JohnFulpWillard:
To be more specific, I've noticed that increasing the size of the game screen makes it even harder to notice when someone is talking to you over radio. It would be best to leave it as-is until fullscreen gets added so chat could be overlayed on top of the game itself.

